PR #5, opened 2022-04-16
Add unit tests for ClickHouse monitor 

yanjunz97:


salv-orlando (dismissed):

yanjunz97:
> These two unit tests overall looks good, since they only cover the monitorMemory() function, are you planning add more tests to cover other functions in this PR? IMO, getDeleteRowNum() also worths testing.

Thanks Yongming for suggestions. Added a test for `getDeleteRowNum` and updated in #13 . This PR is closed as it cannot be updated any more after the repo switched from private to public.

PR #1, opened 2022-04-12
Migrate ClickHouse data retention related contents

yanjunz97:
Signed-off-by: Yanjun Zhou <zhouya@vmware.com>

PR #7, opened 2022-04-26
Add e2e test for FA Grafana

wsquan171:
This is a WIP PR to bring in e2e test for the Flow visibility Grafana stack from antrea.io/antrea repo.

Differences from the main repo e2e test:
- Reduced framework.go to contain only the necessary helper functions
- Removed unused options from TestOptions
- Relocated functions previously under KubernetesUtils to framwork (receiver of TestData), since for FA tests it's only used for policy creation / deletion, and most of pod-facing features are not used.
- Removed IPFIX collector related codes

TBD:
- Bring in runner script and add github workflow when deployment manifest generation PR is merged
- Further cleanup framework.go utility functions
- Validate log and coverage file capturing

To run the test, an existing k8s cluster is required. This can be don for now by using the test scripts in the main repo. The flow aggregator and antrea agent configmap needs to be preset, as we're no longer considering mutating these params as part of test run.


wsquan171:
Thanks @edwardbadboy for the comment. I'll remove the coverage related bits and port in antrea repo e2e updates to this pr tomorrow. #14 has the conflict resolved. 

since many codes are brought over from antrea, current pkg in go.mod (from #1) overlapping with what's used in antrea needs to be downgraded for the time being as we focus on the content of test and the framework in general. the functionality of ch monitor and the build process is not affected.

PR #13, opened 2022-05-04
Add unit tests for ClickHouse monitor

yanjunz97:

### DIFF
--- a/plugins/clickhouse-monitor/main_test.go
+++ b/plugins/clickhouse-monitor/main_test.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"database/sql"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


yanjunz97:
This PR is reopened as the original one #5 cannot be updated any more after the repo switched from private to public.

plugins/clickhouse-monitor/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+	timeRow := sqlmock.NewRows([]string{"timeInserted"}).AddRow(baseTime.Add(5 * time.Second))
+	mock.ExpectQuery("SELECT free_space, total_space FROM system.disks").WillReturnRows(diskRow)
+	mock.ExpectQuery("SELECT COUNT() FROM flows").WillReturnRows(countRow)
+	mock.ExpectQuery("SELECT timeInserted FROM flows LIMIT 1 OFFSET 5").WillReturnRows(timeRow)
+	for _, table := range []string{"flows", "flows_pod_view", "flows_node_view", "flows_policy_view"} {
+		command := fmt.Sprintf("ALTER TABLE %s DELETE WHERE timeInserted < toDateTime('%v')", table, baseTime.Add(5*time.Second).Format(timeFormat))
+		mock.ExpectExec(command).WillReturnResult(sqlmock.NewResult(0, 5))
+	}
+
+	tableName = "flows"
```
heanlan, line 45: May I ask about where do we take the two variables `tableName` `mvNames` in the test?

heanlan (commented):
overall LGTM, just a small question

plugins/clickhouse-monitor/main_test.go
yanjunz97: These two are global variables in monitor code read from the monitor YAML files. Set them as `monitorMemory` will use them.

heanlan: got it, thanks

plugins/clickhouse-monitor/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	monitorMemory(db)
+
+	if err := mock.ExpectationsWereMet(); err != nil {
+		t.Errorf("there were unfulfilled expectations: %s", err)
+	}
+}
+
+func TestGetDeleteRowNum(t *testing.T) {
+	db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
```
dreamtalen, line 73: Since we need to create sqlmock for each test, could we extract it into a function? Or reuse the same db mock connection for all tests in this file.

yanjunz97: Sounds great, updated.

dreamtalen (approved):
LGTM

PR #14, opened 2022-05-05
Add e2e test for FA Grafana

wsquan171:
This is a WIP PR to bring in e2e test for the Flow visibility Grafana stack from antrea.io/antrea repo.

Differences from the main repo e2e test:

Reduced framework.go to contain only the necessary helper functions
Removed unused options from TestOptions
Relocated functions previously under KubernetesUtils to framwork (receiver of TestData), since for FA tests it's only used for policy creation / deletion, and most of pod-facing features are not used.
Removed IPFIX collector related codes
TBD:

Bring in runner script and add github workflow when deployment manifest generation PR is merged
Further cleanup framework.go utility functions
Validate log and coverage file capturing
To run the test, an existing k8s cluster is required. This can be don for now by using the test scripts in the main repo. The flow aggregator and antrea agent configmap needs to be preset, as we're no longer considering mutating these params as part of test run.


wsquan171:
creating a new PR in place of #7 as the publicity of the repo changes. will address @edwardbadboy 's comment in a later patch set.

edwardbadboy (approved):
I have just a minor comment.

test/e2e/fixture.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	log.Println("Checking CoreDNS deployment")
+	if err := data.checkCoreDNSPods(defaultTimeout); err != nil {
+		return err
+	}
+	return nil
+}
+
+func teardownTest(tb testing.TB, data *TestData) {
+	// exportLogs(tb, data, "beforeTeardown", true)
```
salv-orlando, line 32: is this line supposed to be commented out? (sorry for the pedant note, not sure if it's still commented out as the PR is WIP)

salv-orlando (commented):
@edwardbadboy @wsquan171 how much of this code is supposed to be specific for Theia and will evolve independently of Antrea?

I apologise if you already discussed this point, but I wonder if we need to consider the e2e framework as a library which is consumed by Theia (and eventually by Antrea as well) for easier maintanability.

test/e2e/framework.go
```diff
@@ -0,0 +1,1242 @@
...
+	antreaDaemonSet          string = "antrea-agent"
+	antreaDeployment         string = "antrea-controller"
+	flowAggregatorDeployment string = "flow-aggregator"
+	flowAggregatorYML        string = "flow-aggregator.yml"
+	flowAggregatorCovYML     string = "flow-aggregator-coverage.yml"
+	flowVisibilityYML        string = "flow-visibility.yml"
+	flowVisibilityCHPodName  string = "chi-clickhouse-clickhouse-0-0-0"
+
+	agnhostImage  = "k8s.gcr.io/e2e-test-images/agnhost:2.29"
+	busyboxImage  = "projects.registry.vmware.com/library/busybox"
```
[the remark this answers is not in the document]
wsquan171: fixed

test/e2e/fixture.go
wsquan171: fixed. 

Regarding your question above, we had a discussion earlier with @edwardbadboy and decided to start with having an local copy in theia. The issue with importing as lib requires antrea side to export a whole bunch of functions (framework.go, receiver of TestData), which implies 
- a revised abstraction of TestData so the current variants (antrea, multi cluster and theia) still work and make sense
- a better collected of actual tooling methods, which currently scatters all over the place in antrea (under e2e package, but defined in multiple files and certain ones relies on K8s_Utils which is not used here)

I think at this stage we still go with copying related code and trimming it down to only necessary parts, to have a functional test framework to begin with, then come back and decide whether we need to extract some common interface between antrea and theia.

salv-orlando: Understood... I'll make sure to follow up on this :)

salv-orlando (approved):

PR #12, opened 2022-05-04
Add Grafana dashboard for visualizing deny connections

heanlan:
1. Add source files of grafana-chord-plugin
  2. Refactor networkpolicy_dashboard by adding support for visualizing deny
  connections
  3. Add corresponding documentation
  
  Signed-off-by: heanlan <hanlan@vmware.com>

PR #15, opened 2022-05-05
Add Helm support for Theia with ClickHouse PV

yanjunz97:
This PR adds Helm chart for Theia and uses Helm templates instead of Kustomize to generate Theia manifest.


build/charts/theia/values.yaml
```diff
@@ -0,0 +1,27 @@
+clickhouse:
```
antoninbas, line 1: It would make sense to have config options for the different theia images IMO (CH operator, Grafana, etc)

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+  # -- Container image to use for the ClickHouse Monitor.
+  monitorImage:
+    repository: "projects.registry.vmware.com/antrea/theia-clickhouse-monitor"
+    pullPolicy: "IfNotPresent"
+    tag: "latest"
```
antoninbas, line 6: I think something like this would make more sense:
```
clickhouse:
    monitor:
        enable: true
        image:
            repository: ...
            ....
```
there may be more customization later on for the monitor and I think we can also offer the option to disable it?

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+  # -- ClickHouse username. It will be stored in a secret.
+  username: "clickhouse_operator"
+  # -- ClickHouse password. It will be stored in a secret.
+  password: "clickhouse_operator_password"
+  # -- HTTP port number for the ClickHouse service.
+  httpPort: 8123
+  # -- TCP port number for the ClickHouse service.
+  tcpPort: 9000
```
antoninbas, line 14: does it sometimes make sense to disable one of the ports to only use HTTP or TCP?

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+  # -- ClickHouse storage size. Can be a plain integer or as a fixed-point
+  # number using one of these quantity suffixes: E, P, T, G, M, K. Or the
+  # power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki.
+  storageSize: "8Gi"
```
antoninbas, line 18: I was expecting much more customization here (with PV support, etc.)

yanjunz97: I have the plan to have more customization like PV related options and monitor customizations including whether to enable it and some parameters used in the monitor. But I'm not sure whether I should include all of these in this PR or it would be clear to add them in follow-up PRs?

build/charts/theia/values.yaml
yanjunz97: I think the Flow Aggregator uses TCP to connect the ClickHouse while Grafana uses HTTP for connection. Thus it might not have a use case to disable one of them.

yanjunz97:
This PR is converted to draft to add PV supports and some more customizations.

yanjunz97:
Hi @antoninbas I have added the ClickHouse PV support and some more customizations in this PR. I also have several topics regarding Helm support in Theia to bring up. I would like to hear from your options on these topics.

1. **Namespace**
It seems Helm has the convention to use `--namespace` and `--create-namespace` to deploy the chart in a specific Namespace. In the flow visibility case, it works when I use `helm install`. But when I try to generate the manifest with `helm template`, the creation of Namespace will not be included in the generated YAML file even if I use `--create-namespace`. I find here https://github.com/helm/helm/issues/9813 shows Helm team does not plan to provide `--create-namespace` support for `helm template`.
I come up with 2 solutions:

- Use the YAML file without a creation of flow-visibility Namespace, and add instructions to create Namespace `flow-visibility` before applying the `flow-visibility.yml` in docs. But I’m worry about whether it would be a problem as it may bring inconsistency comparing to our previous versions.
- Add a value `createNamespace` to `values.yaml` and set it to true when running the manifest generation script. This can produce a manifest with Namespace as previous. But it seems strange to expose this variable as a value to user.
I’m not sure which solution to choose or whether there is a better solution.

2. **ClickHouse Operator**
ClickHouse Operator is a CRD which should be installed before applying flow-visibility.yml as it defines the kind `ClickHouseInstallation`. I see there is a discussion about whether to put it into the `crds` dir at the issue  https://github.com/antrea-io/antrea/issues/3665. It seems there is no final conclusion on this topic; and ClickHouse Operator is slightly different from CRDs in Antrea:

- The ClickHouse Operator is provided by a third party instead of a CRD developed by us. 
- It seems we cannot treat ClickHouse Operator as a regular templated resource as it must be installed before we try to deploy the ClickHouse server.
I’m not sure whether it is better to put ClickHouse Operator into the `crds` dir?

3. **Release and documentation**
It is mentioned in https://github.com/antrea-io/antrea/pull/3578 that we will have a future PR about the Helm chart release. And I see Helm provides chart-releaser-action at https://github.com/helm/chart-releaser-action. Do we plan to use it or just let user clone our repo first for now? 
There are a lot of instructions on how to modify the YAML file in our doc. IMO these should be removed when we introduce the Helm chart. But before we have the chart in the release, it seems we still need to keep it? Also after we have a chart in release, I’m not sure whether it is replicated to add instructions on how to modify the Helm values when we introduce each features in flow visibility docs, as we already have an auto-generated Helm doc.

also cc @salv-orlando @dreamtalen @heanlan @wsquan171

antoninbas:
@yanjunz97 

1. I know that the Nginx ingress controller uses a kustomize step after rendering the manifest with Helm to inject the `Namespace` resource: https://github.com/kubernetes/ingress-nginx/blob/main/hack/manifest-templates/common/kustomization.yaml. You could consider doing the same thing. Note that in the case of flow visibility, I think we should encourage the user to deploy with Helm directly, instead of applying a manifest `kubectl apply -f`. We should still provide a reference YAML manifest though for users who don't want to use Helm.
2. If you want to include the ClickHouse operator in the Theia Helm chart, you should really put the `ClickHouseInstallation` under the `crds/` directory. Your use case is very different from the Antrea one, given that you actually need to create a CR for that CRD, as part of the Helm deployment. So you should follow Helm's best practices, and ignore the discussion in https://github.com/antrea-io/antrea/issues/3665. BTW, if you package everything as one Helm chart, how do you plan to generate 2 distinct manifests for "manual" installation? If you have a single manifest with both the CRD and the corresponding CR, I think using `kubectl apply` may fail, as the CR may be "created" before the CRD has been registered.
3. We do plan to release the Antrea Helm chart properly (and not require users to clone the repo), but I am trying to figure out https://github.com/antrea-io/antrea/issues/3665 first. We will have a chart repo for all the Antrea charts, including Theia. If Helm becomes your primary installation mechanism for Theia, I think the documentation should reflect that, with instructions on how to modify Helm values to customize the installation.

Makefile
```diff
@@ -123,6 +123,11 @@ verify:
 	@echo "===> Verifying documentation formatting for website <==="
 	$(CURDIR)/hack/verify-docs-for-website.sh
 
+.PHONY: toc
+toc:
+	@echo "===> Generating Table of Contents for Antrea docs <==="
+	GO=$(GO) $(CURDIR)/hack/update-toc.sh
```
salv-orlando, line 129: is this addition related to helm charts?

VERSION
```diff
@@ -1 +1 @@
-v1.7.0-dev
+v0.1.0-dev
```
salv-orlando, line 1: these changes make a lot of sense, but I'd create separate PRs for those

build/charts/theia/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| clickhouse.httpPort | int | `8123` | HTTP port number for the ClickHouse service. |
+| clickhouse.image | object | `{"pullPolicy":"IfNotPresent","repository":"projects.registry.vmware.com/antrea/theia-clickhouse-server","tag":"21.11"}` | Container image to use for the ClickHouse. |
+| clickhouse.monitor.deletePercentage | float | `0.5` | The percentage of records in ClickHouse that will be deleted when the storage grows above threshold. Vary from 0 to 1. |
+| clickhouse.monitor.enable | bool | `true` | Determine whether to run a monitor to periodically check the ClickHouse memory usage and clean data. |
+| clickhouse.monitor.image | object | `{"pullPolicy":"IfNotPresent","repository":"projects.registry.vmware.com/antrea/theia-clickhouse-monitor","tag":"latest"}` | Container image to use for the ClickHouse Monitor. |
```
salv-orlando, line 25: What would be the advantage of making this image configurable? Is it for using different tags?

build/charts/theia/README.md
```diff
@@ -0,0 +1,45 @@
...
+| clickhouse.persistentVolume.affinity | object | `{"required":{"nodeSelectorTerms":[{"matchExpressions":[{"key":"antrea.io/clickhouse-data-node","operator":"Exists"}]}]}}` | Affinity for the Local Persistent Volume. Required when Persistent Volumes is enable and the provisioner is "Local". |
+| clickhouse.persistentVolume.enable | bool | `false` | Enable deploying the ClickHouse with Persistent Volumes. |
+| clickhouse.persistentVolume.localPath | string | `"/data"` | The local path. Required when Persistent Volumes is enable and the provisioner is "Local". |
+| clickhouse.persistentVolume.nfsHost | string | `""` | The NFS server hostname or IP address. Required when Persistent Volumes is enable the provisioner is "NFS". |
+| clickhouse.persistentVolume.nfsPath | string | `""` | The path exported on the NFS server. Required when Persistent Volumes is enable the provisioner is "NFS". |
+| clickhouse.persistentVolume.provisioner | string | `"Local"` | Persistent Volume Provisioner. Required if Persistent Volumes is enable. It must be one of "StorageClass", "Local", "NFS". |
+| clickhouse.persistentVolume.storageClass | string | `""` |  |
```
salv-orlando, line 34: Does this need a description?

build/charts/theia/README.md
```diff
@@ -0,0 +1,45 @@
...
+| clickhouse.storageSize | string | `"8Gi"` | ClickHouse storage size. Can be a plain integer or as a fixed-point number using one of these quantity suffixes: E, P, T, G, M, K. Or the power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki. |
+| clickhouse.tcpPort | int | `9000` | TCP port number for the ClickHouse service. |
```
salv-orlando, line 36: I know it's autogenerated, but is there a way to have this immediately after or before clickhouse.httpPort?

build/charts/theia/README.md
```diff
@@ -0,0 +1,45 @@
...
+| clickhouse.ttl | int | `3600` | Time to live in seconds for data in the ClickHouse. |
+| clickhouse.username | string | `"clickhouse_operator"` | ClickHouse username. It will be stored in a secret. |
```
salv-orlando, line 38: nit: as above, but with clickhouse.password!

build/charts/theia/provisioning/datasources/create_table.sh
```diff
@@ -69,10 +69,10 @@ clickhouse client -n -h 127.0.0.1 <<-EOSQL
         trusted UInt8 DEFAULT 0
     ) engine=MergeTree
     ORDER BY (timeInserted, flowEndSeconds)
-    TTL timeInserted + INTERVAL 1 HOUR
-    SETTINGS merge_with_ttl_timeout = 3600;
+    TTL timeInserted + INTERVAL {{ .Values.clickhouse.ttl }} SECOND
+    SETTINGS merge_with_ttl_timeout = {{ .Values.clickhouse.ttl }};
 
-    CREATE MATERIALIZED VIEW flows_pod_view
+    CREATE MATERIALIZED VIEW IF NOT EXISTS flows_pod_view
```
salv-orlando, line 75: This also seems a change not related to helm charts.

docs/network-flow-visibility.md
```diff
@@ -56,14 +56,29 @@ ClickHouse as the data storage, and use Grafana as the data visualization and mo
 
 ### Deployment Steps
 
-To deploy the Grafana Flow Collector, the first step is to install the ClickHouse
-Operator, which creates, configures and manages ClickHouse clusters. Check the [homepage](https://github.com/Altinity/clickhouse-operator)
-for more information about the ClickHouse Operator. Current checked-in yaml is based on their
-[v0.18.2](https://github.com/Altinity/clickhouse-operator/blob/refs/tags/0.18.2/deploy/operator/clickhouse-operator-install-bundle.yaml) released version. Running the following command
-will install ClickHouse Operator into `kube-system` Namespace.
+From Theia v0.1, we support deploying the Grafana Flow Collector with Helm. Here
```
salv-orlando, line 59: Since nothing was released before 0.1 for theia we don't need to specify "from theia 0.1"

docs/network-flow-visibility.md
```diff
@@ -297,11 +326,161 @@ storage size, please modify the `sizeLimit` in the following section.
...
 
+To deploy ClickHouse with Persistent Volumes and limited storage size, please refer
+to [Persistent Volumes](#persistent-volumes).
+
 The time interval between the batch commits to the ClickHouse is specified in the
 [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
 as `commitInterval`. The ClickHouse throughput grows sightly when the commit interval
 grows from 1s to 8s. A commit interval larger than 8s provides little improvement on the throughput.
 
+##### Persistent Volumes
```
salv-orlando, line 337: wasn't this already in your PR for persistent volumes?

hack/generate-manifest.sh
```diff
@@ -96,36 +100,32 @@ elif ! $KUSTOMIZE version > /dev/null 2>&1; then
...
+    EXTRA_VALUES="--set clickhouse.monitorImage.repository=$IMG_NAME,clickhouse.monitorImage.tag=$IMG_TAG"
 fi
 
+THEIA_CHART=$THIS_DIR/../build/charts/theia
+KUSTOMIZATION_DIR=$THIS_DIR/../build/yamls
+# intermediate manifest
+MANIFEST=$KUSTOMIZATION_DIR/base/manifest.yaml
+# Suppress potential Helm warnings about invalid permissions for Kubeconfig file
+# by throwing away related warnings.
+$HELM template \
```
salv-orlando, line 116: I would like to understand if we can completely drop kustomize by using helm --create-namespace with https://helm.sh/docs/faq/changes_since_helm2/#automatically-creating-namespaces

Would there be any issue? for instance, would helm fail if the namespace already exist?

salv-orlando (commented):
There seem to be some changes not related to helm charts - can we have distinct PRs for those?

Makefile
yanjunz97: No. Migrating this to make toc generation more convenient. Put it in a separate PR https://github.com/antrea-io/theia/pull/21 Thanks!

VERSION
yanjunz97: Thanks! I remove it from this PR.

build/charts/theia/README.md
yanjunz97: I think it brings more flexibility to config the image tag and pullPolicy.

build/charts/theia/README.md
yanjunz97: It is missed. Added.

build/charts/theia/README.md
yanjunz97: It makes sense! I use `clickhouse.port.http` and `clickhouse.port.tcp` make them grouped together.

build/charts/theia/README.md
yanjunz97: Use `clickhouse.secret` to group these two.

build/charts/theia/provisioning/datasources/create_table.sh
yanjunz97: This is related to the ClickHouse PV support. PV support is also added into this PR as

- There are only a few customization variables before we add PV support
- PV support contents have been reviewed at Antrea side. Adding PV to this PR mainly requires reviews on the Helm related contents.

docs/network-flow-visibility.md
yanjunz97: Removed "from theia 0.1".

docs/network-flow-visibility.md
yanjunz97: As explained above https://github.com/antrea-io/theia/pull/15#discussion_r870653618, PV is also included in this PR now.

hack/generate-manifest.sh
yanjunz97: We support using `--create-namespace` when deploying by `helm install`. But we still keep the script for generating a standard `flow-visibility.yaml`. Kustomize is necessary here as `helm template` does not support `--create-namespace`.

yanjunz97:
Thanks @antoninbas ! This PR is updated and ready for review now. Here is the update on the topic we have discussed above.

1. I have updated the scripts to use Kustomize to inject the Namespace.
2. For the ClickHouse Operator, I put it into the `crds` dir. It seems if we do not add the `--include-crds` option when running `helm template`, the contents in `crds` dir will not be included into the generated YAML file. Thus we still have two distinct yaml files, one in the `crds` dir and the `flow-visibility.yml` generated by the script. Users can still use the same way as before to manually deploy the flow visibility contents.
3. The instructions on how to modify the YAML file will be replaced by Helm instructions in a future PR when we release the Helm chart.

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,79 @@
...
+    # -- HTTP port number for the ClickHouse service.
+    http: 8123
+    # -- TCP port number for the ClickHouse service.
+    tcp: 9000
+  # -- ClickHouse storage size. Can be a plain integer or as a fixed-point
+  # number using one of these quantity suffixes: E, P, T, G, M, K. Or the
+  # power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki.
+  storageSize: "8Gi"
+  # -- Time to live in seconds for data in the ClickHouse.
+  ttl: 3600
```
wsquan171, line 37: as PV is introduced, I think data is expected to be kept with longer lifespan? counting ttl in seconds maybe turn out to be inconvenient. should we make it a string and support other type of `Intervals` (i.e. HOUR / DAY / WEEK etc.?)

docs/network-flow-visibility.md
```diff
@@ -56,14 +59,29 @@ ClickHouse as the data storage, and use Grafana as the data visualization and mo
...
+to customize the installation.
+
+You can clone the repository and run the following command to install the Grafana
+Flow Collector into Namespace `flow-visibility`.
+
+```bash
+helm install -f <path-to-your-values-file> theia ./build/charts/theia -n flow-visibility --create-namespace
+```
+
+We recommend using Helm to deploy the Grafana Flow Collector. But if you prefer
```
wsquan171, line 74: since e2e tests not longer mutates configMap as part of the run, do you think we should also switch to helm in jenkins and kind tests? also, maybe add a separate e2e job for local PV?

wsquan171:
> 2. For the ClickHouse Operator, I put it into the `crds` dir. It seems if we do not add the `--include-crds` option when running `helm template`, the contents in `crds` dir will not be included into the generated YAML file. Thus we still have two distinct yaml files, one in the `crds` dir and the `flow-visibility.yml` generated by the script. Users can still use the same way as before to manually deploy the flow visibility contents.

I kind of think that moving ch op install bundle into crd sub dir is a bit hacky. it's not technically just a crd yml, and if in the future we need to introduce crd for Theia we'll eventually have to use the --include-crds switch. is the main goal of putting it under crd for having one single `helm install` command?

build/yamls/flow-visibility.yml
```diff
@@ -4681,12 +4682,12 @@ data:
...
 kind: ConfigMap
 metadata:
-  name: grafana-dashboard-config-h4fg25d7k9
+  name: grafana-dashboard-config
   namespace: flow-visibility
 ---
 apiVersion: v1
 data:
-  dashboard_provider.yml: |
+  dashboard_provider.yaml: |-
```
dreamtalen, line 4690: Is this name change from `.yml` to `.yaml` is required by Helm?

ci/kind/test-e2e-kind.sh
```diff
@@ -151,7 +151,7 @@ function run_test {
 
   docker exec -i kind-control-plane dd of=/root/antrea.yml < $TMP_DIR/antrea.yml
   docker exec -i kind-control-plane dd of=/root/flow-aggregator.yml < $TMP_DIR/flow-aggregator.yml
-  docker exec -i kind-control-plane dd of=/root/clickhouse-operator-install-bundle.yml < $CH_OPERATOR_YML
+  docker exec -i kind-control-plane dd of=/root/clickhouse-operator-install-bundle.yaml < $CH_OPERATOR_YML
```
dreamtalen, line 154: Same question above, `.yml` changed to `.yaml`.

dreamtalen, line 154: Same question above, `.yml` changed to `.yaml`.

build/yamls/flow-visibility.yml
yanjunz97: it is not mandatary by Helm. Helm only requires files like `Chart.yaml` and `values.yaml` to be ended by `.yaml`. I change all the YAML files extension under Helm chart dir to `.yaml` to keep consistency and avoid confusions or errors in future. I can change some of them back if you think this do bring new confusions.

dreamtalen: Got it, current change sounds good to me for consistency.

yanjunz97:
> > 2. For the ClickHouse Operator, I put it into the `crds` dir. It seems if we do not add the `--include-crds` option when running `helm template`, the contents in `crds` dir will not be included into the generated YAML file. Thus we still have two distinct yaml files, one in the `crds` dir and the `flow-visibility.yml` generated by the script. Users can still use the same way as before to manually deploy the flow visibility contents.
> 
> I kind of think that moving ch op install bundle into crd sub dir is a bit hacky. it's not technically just a crd yml, and if in the future we need to introduce crd for Theia we'll eventually have to use the --include-crds switch. is the main goal of putting it under crd for having one single `helm install` command?

Thanks @wsquan171 for pointing out this. Yes, by putting it under `crds`, we can support a single `helm install` command for the whole product. As stated in [Helm docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/), I think the purpose of `crds` dir is mainly for the installation of CRD declaration before using the resource, which is pretty much similar to our case IMO. And I'm worry about it would be confusing if users have to use `kubectl apply` and `helm install` at the same time to deploy Theia.

But I also understand your concern about what if we add our own CRDs. I think these CRDs are under the similar conditions mentioned in the Antrea issue https://github.com/antrea-io/antrea/issues/3665 . These CRDs are kept as Template in Antrea chart now.

Basically there are two strategies to treat the ClickHouse Operator:

- Remove ClickHouse Operator from Theia Chart and state it as a prerequisite of Theia.
- Keep as it is and revisit the topic when we have our own CRDs and there is an update of the placement CRDs on the Antrea chart.

I'd also like to have @antoninbas option on this.

docs/network-flow-visibility.md
yanjunz97: In jenkins and kind tests, I think we can still use the YAML file and manifest generating scripts as Antrea side does now. As the manifest generating scripts are updated to use the `helm template` and it is easier to override values.yaml with the script. Adding a separate e2e job for local PV sounds feasible for me.

docs/network-flow-visibility.md
```diff
@@ -230,6 +258,8 @@ Collector in production.
 
 #### ClickHouse Configuration
 
+##### Service Customization
```
heanlan, line 261: A general comment on all of the original configuration instructions: If we keep all these original configuration instructions for users who don't to use Helm, can we write it explicitly in the doc, with something like: for all of the configurations, option 1 - with Helm, edit values.yaml, option 2 - without Helm, edit flow-visibility.yml directly.

Otherwise it seems a bit confusing to me, if we recommend using Helm to template yaml, and we also recommend making a series of configurations directly on flow-visibility.yml in the meanwhile.

Makefile
```diff
@@ -157,4 +157,3 @@ clickhouse-monitor:
 clickhouse-monitor-plugin:
 	@mkdir -p $(BINDIR)
 	GOOS=linux $(GO) build -o $(BINDIR) $(GOFLAGS) -ldflags '$(LDFLAGS)' antrea.io/theia/plugins/clickhouse-monitor
-
```
heanlan, line 160: new line

build/charts/theia/templates/grafana/deployment.yaml
```diff
@@ -103,8 +21,8 @@ spec:
           - 0
       containers:
         - name: grafana
-          image: projects.registry.vmware.com/antrea/theia-grafana:8.3.3
-          imagePullPolicy: IfNotPresent
+          image: {{ .Values.grafana.image.repository }}:{{ .Values.grafana.image.tag }}
+          imagePullPolicy: {{ .Values.grafana.image.pullPolicy }}
           env:
             - name: GF_INSTALL_PLUGINS
               value: "https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.0.zip;theia-grafana-sankey-plugin,grafana-clickhouse-datasource 1.0.1"
```
heanlan, line 28: shall we add the value of `GF_INSTALL_PLUGINS` to values.yaml?

build/charts/theia/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Source Code
+
+* <https://github.com/antrea-io/theia>
+
+## Requirements
+
+Kubernetes: `>= 1.16.0-0`
+
+## Values
```
heanlan, line 17: to decrease scrolling right, shall we switch the Default and Description column? like https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#configuration

build/charts/theia/templates/grafana/dashboard-configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: grafana-dashboard-config
+  namespace: {{ .Release.Namespace }}
+data:
+{{ (.Files.Glob "provisioning/dashboards/*.json").AsConfig | indent 2}}
```
heanlan, line 7: Is that possible to add the dashboard json file names as a parameter into values.yaml? so that users can select on which dashboards they want to include in Grafana

build/charts/theia/values.yaml
yanjunz97: Added

build/charts/theia/values.yaml
yanjunz97: Updated.

docs/network-flow-visibility.md
yanjunz97: I make an update on the document, hope it clearer now.

build/charts/theia/templates/grafana/deployment.yaml
yanjunz97: Added.

build/charts/theia/README.md
yanjunz97: This doc is auto-generated by [helm-docs](https://github.com/norwoodj/helm-docs/releases/v1.7.0) as Antrea side does. I do not find a way switch the column. I can revisit the problem if the doc generation way is updated in the future.

build/charts/theia/templates/grafana/dashboard-configmap.yaml
yanjunz97: Updated.

Makefile
yanjunz97: I think it means the old file has no newline at end of file, as the last line have 2 spaces on the final line. I just removed that 2 spaces in this commit.

build/charts/theia/values.yaml
yanjunz97: Thanks, it is supported now!

yanjunz97:
Hi @antoninbas Could you take another look at this PR?

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    httpPort: 8123
+  # -- ClickHouse storage size. Can be a plain integer or as a fixed-point
+  # number using one of these quantity suffixes: E, P, T, G, M, K. Or the
+  # power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki.
+  storageSize: "8Gi"
+  # -- Time to live for data in the ClickHouse. Can be a plain integer using
+  # one of these unit suffixes SECOND, MINUTE, HOUR, DAY, WEEK, MONTH, QUARTER,
+  # YEAR.
+  ttl: 1 HOUR
+  persistentVolume:
```
antoninbas, line 48: I think this whole section can be a bit confusing, because it conflates some things, like PV type & storage class provisioner, and doesn't clearly distinguish between manual volume provisioning and dynamic volume provisioning.

It seems that there is also no option for users to provide a custom PersistentVolume (if they don't want to use dynamic provisioning).

```yaml
storage:
  size: "8Gi"
  createPersistentVolume:
    # can be set to "Local" or "NFS"
    type: ""
    local:
      path: "/data"
    nfs:
      host: ""
      path: ""
  # ignored if persistentVolume.type is non-empty
  persistentVolumeClaimSpec: {}
    # storageClassName: ""
    # volumeName: 
```

By default, memory storage is used.

To use a local volume created by us, one can set:
```
storage.createPersistentVolume.type = "Local"
```

To use a custom PersistentVolume, one can set:
```
storage.persistentVolumeClaimSpec:
  storageClassName: ""
  volumeName: "<my-pv>"
```

To dynamically provision a PersistentVolume, one can set:
```
storage.persistentVolumeClaimSpec
  storageClassName: <"my-storage-class">
```

Let me know what you think of that solution

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    # -- Enable deploying the ClickHouse with Persistent Volumes.
+    enable: false
+    # -- Persistent Volume Provisioner. Required if Persistent Volumes is enable.
```
antoninbas, line 51: is enabled

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    # It must be one of "StorageClass", "Local", "NFS".
+    provisioner: "Local"
+    # -- The local path. Required when Persistent Volumes is enable and the
```
antoninbas, line 54: is enabled

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    # is enable the provisioner is "NFS".
+    nfsHost: ""
+    # -- The path exported on the NFS server. Required when Persistent Volumes
+    # is enable the provisioner is "NFS".
+    nfsPath: ""
+    # -- The StorageClass used to dynamically provision the Persistent Volume.
+    # Required when Persistent Volumes is enable the provisioner is "StorageClass".
+    storageClass: ""
+grafana:
+  # -- Container image to use for the Grafana.
```
antoninbas, line 76:
```suggestion
  # -- Container image to use for Grafana.
```

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    repository: "projects.registry.vmware.com/antrea/theia-grafana"
+    pullPolicy: "IfNotPresent"
+    tag: "8.3.3"
+  secret:
+    # -- Grafana username. It will be stored in a secret.
+    username: "admin"
+    # -- Grafana password. It will be stored in a secret.
+    password: "admin"
+  service:
+    # -- The type of service exposes Grafana. It must be one of NodePort or LoadBalancer.
```
antoninbas, line 87:
```suggestion
    # -- The type of service exposing Grafana. It must be one of NodePort or LoadBalancer.
```

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    type: NodePort
+    # -- TCP port number for the Grafana service.
+    tcpPort: 3000
+  # -- Grafana plugins to install.
+  installPlugins: "https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.0.zip;theia-grafana-sankey-plugin,grafana-clickhouse-datasource 1.0.1"
```
antoninbas, line 92: maybe it would be more user-friendly to support a YAML array here. You can then build the string when rendering the template.

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    # -- The NFS server hostname or IP address. Required when Persistent Volumes
+    # is enable the provisioner is "NFS".
+    nfsHost: ""
+    # -- The path exported on the NFS server. Required when Persistent Volumes
+    # is enable the provisioner is "NFS".
+    nfsPath: ""
+    # -- The StorageClass used to dynamically provision the Persistent Volume.
+    # Required when Persistent Volumes is enable the provisioner is "StorageClass".
+    storageClass: ""
+grafana:
```
antoninbas, line 75: have we considered the case where a user is already running an instance of Grafana and want to use the same instance for flow visibility? @salv-orlando

build/charts/theia/values.yaml
yanjunz97: I think PersistentVolumeClaim does not accept PV name as a specification. As introduced in [K8S doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1), I think it means

- If storageClassName is specified, PVC will look for PV in the cluster with that class
- If storageClassName is set to "", PVC will look for PV with no class

From my experience, if users don't want to use dynamic provisioning, they could set StorageClass to "" or create a StorageClass with `kubernetes.io/no-provisioner` as provisioner, and make sure the corresponding PV is created manually. 

We can still adopt the solution, with the `storage.persistentVolumeClaimSpec` only expecting `storageClassName`. Memory storage will be used when `storage. createPersistentVolume.type` is "" and `storage.persistentVolumeClaimSpec` is empty. But I'm not sure whether this change is necessary as it does not distinguish between manual volume provisioning and dynamic volume provisioning either.

build/charts/theia/values.yaml
yanjunz97: I think we do not have a design for this before. To my knowledge, it is possible to connect an existing Grafana to ClickHouse database by setting the service IP and credentials in the Grafana UI. 

But If the existing instance of Grafana is out of cluster, it might have trouble to connect to ClickHouse as ClickHouse is exposed with a ClusterIP service.

And it seems users have to take care of the dashboards import and plugins installation by themselves. I'm not sure whether it is convenient for the users to do these. 

Also would like to include @heanlan in this discussion. Please correct me if I'm wrong.

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+      repository: "projects.registry.vmware.com/antrea/theia-clickhouse-monitor"
+      pullPolicy: "IfNotPresent"
+      tag: "latest"
+  secret:
+    # -- ClickHouse username. It will be stored in a secret.
+    username: "clickhouse_operator"
+    # -- ClickHouse password. It will be stored in a secret.
+    password: "clickhouse_operator_password"
+  service:
+    # -- The type of service exposes ClickHouse. It can be one of ClusterIP, NodePort or LoadBalancer.
```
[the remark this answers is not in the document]
yanjunz97: Updated.

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
+clickhouse:
+  # -- Container image to use for the ClickHouse.
```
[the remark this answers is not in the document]
yanjunz97: Updated

build/charts/theia/values.yaml
yanjunz97: Using the YAML array now, thanks!

build/charts/theia/values.yaml
antoninbas: There are 3 cases for storageClassName: set, omitted and set to empty
I do think it is possible to manually bind to a PV with the `volumeName` field: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume

yanjunz97: I missed this. Thanks! The storage section is refactored now.

build/charts/theia/README.md
salv-orlando: Make sense, thanks!

docs/network-flow-visibility.md
salv-orlando: Sounds good...

build/charts/theia/values.yaml
salv-orlando: @antoninbas I think the answer is that so far we have only been considering a scenario where antrea network flow visibility has its own grafana instance.

But to the best of my knowledge there is nothing special in the way we deploy grafana. In theory we could have users leverage their own grafana instance. I would not know however how to automate dashboard provisioning with helm in this case. This is something we can considering for a future release.

build/charts/theia/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: theia
+type: application
+displayName: Theia
+home: https://antrea.io/
+version: 0.1.0-dev
+appVersion: 0.1.0-dev
```
salv-orlando, line 7: I think we need to update these before we release?

build/charts/theia/provisioning/dashboards/pod_to_external_dashboard.json
```diff
@@ -60,10 +60,10 @@
...
-                    "rawSql": "select SUM(octetDeltaCount) as bytes, sourcePodName as source, destinationIP as destination\nFrom flows_pod_view\nWHERE flowType == 3\nAND sourcePodNamespace NOT IN ('kube-system', 'flow-visibility', 'flow-aggregator')\nAND $__timeFilter(flowEndSeconds)\nGROUP BY source, destination\nHAVING bytes != 0",
+                    "rawQuery": "SELECT SUM(octetDeltaCount), (sourcePodName, destinationIP) AS pair\nFROM default.flows_pod_view\nWHERE flowEndSeconds >= toDateTime(1642534343) AND flowEndSeconds <= toDateTime(1642536143)\nAND flowType == 3\nAND sourcePodNamespace NOT IN ('kube-system', 'flow-aggregator', '{{ .Release.Namespace }}')\nGROUP BY pair",
+                    "rawSql": "select SUM(octetDeltaCount) as bytes, sourcePodName as source, destinationIP as destination\nFrom flows_pod_view\nWHERE flowType == 3\nAND sourcePodNamespace NOT IN ('kube-system', '{{ .Release.Namespace }}', 'flow-aggregator')\nAND $__timeFilter(flowEndSeconds)\nGROUP BY source, destination\nHAVING bytes != 0",
```
salv-orlando, line 66: As PR #12  is making some changes to query, we might want to merge that one, rebase this PR, and then make sure `{{ .Release.Namespace }} ` is used  in every query in place of flow-visibility

salv-orlando (commented):
Code LGTM.

Due to operations done in this PR on Grafana's clickhouse queries - mostly replacing namespace name with variable, I suggest to merge PR #12 , and then rebase this one.

build/charts/theia/Chart.yaml
yanjunz97: Yes. I think we will need to update these in the release branch.

build/charts/theia/values.yaml
antoninbas: I understand that this is not something we need to look at in details now. I just wanted to make sure that this is something that we will be able to accomodate without any significant change to the Helm values file. For example, we should be able to easily disable the deployment of Grafana (e.g., by setting `grafana.enable` to `false`). I don't know if anything else would be required that may have an impact on the structure of values.yaml.

yanjunz97: As ClickHouse does not have dependency on Grafana, I think disabling Grafana will not require a significant change to the Helm values file. There might require some extra steps for user to update their own Grafana instance to display the dashboards, but these should not be related to our Helm chart.

build/charts/theia/README.md
```diff
@@ -0,0 +1,50 @@
...
+| clickhouse.storage.createPersistentVolume.type | string | `""` | Type of PersistentVolume. Can be set to "Local" or "NFS". Please set this value to use a PersistentVolume created by Theia. |
+| clickhouse.storage.persistentVolumeClaimSpec | string | `nil` | Specification for PersistentVolumeClaim. This is ignored if createPersistentVolume.type is non-empty.  To use a custom PersistentVolume, please set  storageClassName: "" volumeName: "<my-pv>" To dynamically provision a PersistentVolume, please set storageClassName: "<my-storage-class>" Memory storage is used if both createPersistentVolume.type and persistentVolumeClaimSpec are empty. |
+| clickhouse.storage.size | string | `"8Gi"` | ClickHouse storage size. Can be a plain integer or as a fixed-point number using one of these quantity suffixes: E, P, T, G, M, K. Or the power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki. |
+| clickhouse.ttl | string | `"1 HOUR"` | Time to live for data in the ClickHouse. Can be a plain integer using one of these unit suffixes SECOND, MINUTE, HOUR, DAY, WEEK, MONTH, QUARTER, YEAR. |
```
ziyouw, line 40: Should we support small intervals, such as second? Seems it is hard to do anything if the ttl is set at second level.

build/charts/theia/provisioning/datasources/create_table.sh
```diff
@@ -14,6 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{- $ttl := split " " .Values.clickhouse.ttl }}
```
ziyouw, line 17: For DAY, WEEK, MONTH, QUARTER, YEAR, we do not have any mapping here?

build/charts/theia/README.md
yanjunz97: These are units supported by ClickHouse. I think keeping it may be useful for users who prefer to use the small units like 3600 SECOND instead of 1 HOUR.

yanjunz97:
> Code LGTM.
> 
> Due to operations done in this PR on Grafana's clickhouse queries - mostly replacing namespace name with variable, I suggest to merge PR #12 , and then rebase this one.

Thanks @salv-orlando This PR is rebased now. Would you like to take another look?

hack/generate-manifest.sh
```diff
@@ -96,36 +100,32 @@ elif ! $KUSTOMIZE version > /dev/null 2>&1; then
...
-touch kustomization.yml
-$KUSTOMIZE edit add base $BASE
-# ../../patches/$MODE may be empty so we use find and not simply cp
-find ../../patches/$MODE -name \*.yml -exec cp {} . \;
-
-if [ "$MODE" == "dev" ]; then
-    $KUSTOMIZE edit set image clickhouse-monitor=projects.registry.vmware.com/antrea/theia-clickhouse-monitor:latest
-    $KUSTOMIZE edit add patch --path imagePullPolicy.yml --group clickhouse.altinity.com --version v1 --kind ClickHouseInstallation --name clickhouse
-fi
+HELM_TMP_DIR=$(mktemp -d $THIS_DIR/../build/yamls/chart-values.XXXXXXXX)
```
antoninbas, line 103: you don't seem to be using this anywhere?

build/yamls/base/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: flow-visibility
+
```
antoninbas, line 4: nit: why the empty line?

hack/generate-manifest.sh
```diff
@@ -96,36 +100,32 @@ elif ! $KUSTOMIZE version > /dev/null 2>&1; then
...
+# Suppress potential Helm warnings about invalid permissions for Kubeconfig file
+# by throwing away related warnings.
+$HELM template \
+      --namespace flow-visibility \
+      $EXTRA_VALUES \
+      "$THEIA_CHART"\
+      2> >(grep -v 'This is insecure' >&2)\
+      > $MANIFEST
+
+# Add flow-visibility Namespace resource by Kustomize
```
antoninbas, line 123:
```suggestion
# Add flow-visibility Namespace resource with Kustomize
```

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,103 @@
...
+    # -- Container image to use for the ClickHouse Monitor.
+    image:
+      repository: "projects.registry.vmware.com/antrea/theia-clickhouse-monitor"
+      pullPolicy: "IfNotPresent"
+      tag: "latest"
+  secret:
+    # -- ClickHouse username. It will be stored in a secret.
+    username: "clickhouse_operator"
+    # -- ClickHouse password. It will be stored in a secret.
+    password: "clickhouse_operator_password"
```
antoninbas, line 32: I assume that these 2 fields will be stored in the same secret, and not 2 separate secrets?
I suggest changing this to:
```
# -- Credentials to connect to ClickHouse. They will be stored in a secret.
connectionSecret:
    username : "clickhouse_operator"
    password: "clickhouse_operator_password"
```

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,103 @@
...
+  # -- Container image to use for Grafana.
+  image: 
+    repository: "projects.registry.vmware.com/antrea/theia-grafana"
+    pullPolicy: "IfNotPresent"
+    tag: "8.3.3"
+  secret:
+    # -- Grafana username. It will be stored in a secret.
+    username: "admin"
+    # -- Grafana password. It will be stored in a secret.
+    password: "admin"
```
antoninbas, line 84: same comment as above

antoninbas (commented):
some minor comments, otherwise LGTM

build/yamls/base/kustomization.yaml
yanjunz97: Removed.

build/charts/theia/values.yaml
yanjunz97: Updated.

build/charts/theia/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    persistentVolumeClaimSpec: 
+      # storageClassName: ""
+      # volumeName: ""
+grafana:
+  # -- Container image to use for Grafana.
+  image: 
+    repository: "projects.registry.vmware.com/antrea/theia-grafana"
+    pullPolicy: "IfNotPresent"
+    tag: "8.3.3"
+  # -- Credentials to login Grafana. They will be stored in a secret.
```
antoninbas, line 79: login to Grafana

antoninbas (approved):

PR #25, opened 2022-05-20
 Implementation of theia and policy recommendation CLI

dreamtalen:
In this commit, we add the implementation of theia, the command-line
tool for Theia based on cobra framework.
Policy recommendation CLI is also included. It consists of 3 commands
which let users could run a new policy recommendation job, check the
current status of a previous job, and retrieve the recommendation result of a
previous job respectively.


Makefile
```diff
@@ -157,4 +157,14 @@ clickhouse-monitor:
 clickhouse-monitor-plugin:
 	@mkdir -p $(BINDIR)
 	GOOS=linux $(GO) build -o $(BINDIR) $(GOFLAGS) -ldflags '$(LDFLAGS)' antrea.io/theia/plugins/clickhouse-monitor
-	
\ No newline at end of file
+
+.PHONY: theiactl
```
salv-orlando, line 161: I think we also need to modify githubactions to ensure the CLI is compiled and shipped with release assets

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+because policy recommendation feature needs Pod lables information to recommend network policies.
+
+Then, please follow these [deployment steps](network-flow-visibility.md#deployment-steps)
+to deploy the Grafana Flow Collector.
+
+After we finish the deployment of Grafana Flow Collector and can see some flow records in the Grafana UI,
+we can continue to deploy other necessary components of policy recommendation feature:
+
+We need to install the [Kubernetes Operator for Apache Spark](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator)
+in the cluster.
```
salv-orlando, line 36: @yanjunz97  - Can you assess how difficult would it to add this to helm charts as well?
If not possible, we might want to document that the Theia helm chart does not yet suppport Spark Operator deployment

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+NAME                                          READY   STATUS    RESTARTS   AGE
+policy-reco-spark-operator-56c4cb454c-4vhfh   1/1     Running   0          2m19s
+```
+
+## Command-line Usage
+
+Currently, the network policy recommendation feature only supports command line interaction through Theiactl.
+Theiactl is the Theia command-line tool. To get more information about Theiactl, please refer to this [doc](./theiactl.md). We have 3 Theiactl commands for the policy recommendation feature:
+
+- `theiactl policyreco start`
```
salv-orlando, line 62: naming question: I am thinking recommend-policy might be a better name, but on the other hand it's also quite long... The thing about "reco" is that it might be associated to terms like "recognition" or "reconnaissance"

Thoughts?

pkg/theiactl/commands/root.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "theiactl",
+	Short: "theiactl is the command line tool for Theia",
+	Long: `theiactl is the command line tool for Theia that supports the
+policy recommendation feature`,
```
salv-orlando, line 28: For the time being I'd keep "short" and "long" description the same or change it to

`theiactl is the command line tool for Theia which provides access to some network flow visibility capabilities`

I don't think we want to expand the long description for every feature we add.

pkg/theiactl/commands/root.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
```
salv-orlando, line 32: Do we have main.main() here? :)
The comment lines can probably be removed.

pkg/theiactl/commands/start.go
```diff
@@ -0,0 +1,378 @@
+// Copyright 2022 Antrea Authors
```
salv-orlando, line 1: maybe we should prefix these module names with "policyreco" as in the future we might have "start" command for many jobs.

pkg/theiactl/commands/start.go
```diff
@@ -0,0 +1,378 @@
...
+		if limit < 0 {
+			return fmt.Errorf("limit should be an integer >= 0")
+		}
+		recoJobArgs = append(recoJobArgs, "--limit", strconv.Itoa(limit))
+
+		option, err := cmd.Flags().GetInt("option")
+		if err != nil {
+			return err
+		}
+		if option < 1 || option > 3 {
```
salv-orlando, line 93: I would prefer to map these integer values to options that ore more intelligible.

For instance instead of --option 3 can user specify --use-k8s-net-pol?

pkg/theiactl/commands/start.go
```diff
@@ -0,0 +1,378 @@
...
+			return err
+		}
+		if endTime != "" {
+			_, err = time.Parse("2006-01-02 15:04:05", endTime)
+			if err != nil {
+				return fmt.Errorf(`parsing end_time: %v, end_time should be in 
+'YYYY-MM-DD hh:mm:ss' format, for example: 2006-01-02 15:04:05`, err)
+			}
+			recoJobArgs = append(recoJobArgs, "--end_time", endTime)
+		}
```
salv-orlando, line 122: Can we add a simple check to make sure that if both start and end time are specified start_time < end_time ?

pkg/theiactl/commands/start.go
```diff
@@ -0,0 +1,378 @@
...
+		nsAllowList, err := cmd.Flags().GetString("ns_allow_list")
+		if err != nil {
+			return err
+		}
+		if nsAllowList != "" {
+			var parsedNsAllowList []string
+			err := json.Unmarshal([]byte(nsAllowList), &parsedNsAllowList)
+			if err != nil {
+				return fmt.Errorf(`parsing ns_allow_list: %v, ns_allow_list should 
+be a list of namespace string, for example: '["kube-system","flow-aggregator","flow-visibility"]'`, err)
```
salv-orlando, line 133: Would it be too difficult for the CLI to parse a list of comma-separated strings? The  JSON notation is not super user-friendly

pkg/theiactl/commands/start.go
```diff
@@ -0,0 +1,378 @@
...
+	)
+	startCmd.Flags().StringP(
+		"end_time",
+		"e",
+		"",
+		`The end time of the flow records considered for the policy recommendation.
+Format is YYYY-MM-DD hh:mm:ss in UTC timezone. No limit of the end time of flow records by default.`,
+	)
+	startCmd.Flags().StringP(
+		"ns_allow_list",
```
salv-orlando, line 330: In K8S' world don't CLI option use a dash as a separator?

salv-orlando (commented):
Initial review

heanlan (commented):
A few nits while looking through the doc

yanjunz97:
I added a commit to implement the two CLI commands to

- check the current status of a previous job
- get the recommendation result of a previous job

Currently `kubectl port-forward` is used for port forwarding. I'm still working on the unit tests for these two commands. Will update the commit once they are finished.

docs/network-policy-recommendation.md
yanjunz97: It should not be difficult to add it. I would take a look at it.

pkg/theiactl/commands/check.go
```diff
@@ -72,12 +78,88 @@ $ theiactl policyreco check --id e998433e-accb-4888-9fc8-06563f073e86
...
+	json.Unmarshal([]byte(response), &getAppsResult)
+	if len(getAppsResult) != 1 {
+		return "", fmt.Errorf("wrong number of Spark Application with id %s, expected 1, got %d", id, len(getAppsResult))
+	}
+	sparkAppID := getAppsResult[0]["id"]
+	// Check the percentage of completed stages
+	url = fmt.Sprintf("http://localhost:4040/api/v1/applications/%s/stages", sparkAppID)
+	response, err = getResponseFromSparkMonitoringSvc(url)
+	if err != nil {
+		return "", fmt.Errorf("failed to get response from Spark Monitoring service at %s, %v", url, err)
```
salv-orlando, line 117: nit about error message. The URL is probably meaningless to the user.
For the error message it might make sense to return the job id.

pkg/theiactl/commands/check.go
```diff
@@ -72,12 +78,88 @@ $ theiactl policyreco check --id e998433e-accb-4888-9fc8-06563f073e86
...
+			}
+			state += stateProgress
+		}
 		fmt.Printf("Status of this policy recommendation job is %s\n", state)
 		return nil
-		// TODO: add implementation of checking work progress through Spark Monitoring Service after port forwarder finished
 	},
 }
 
+func getPolicyRecommendationProgress(id string) (string, error) {
```
salv-orlando, line 100: is the 'id' passed to this function the same as the SparkAppId that would be retrieved from the monitoring service?
If so, do we actually need to pass it?

salv-orlando, line 100: is the 'id' passed to this function the same as the SparkAppId that would be retrieved later on?
If so, do we actually need to pass it?

pkg/theiactl/commands/check.go
```diff
@@ -72,12 +78,88 @@ $ theiactl policyreco check --id e998433e-accb-4888-9fc8-06563f073e86
...
+	// Check the percentage of completed stages
+	url = fmt.Sprintf("http://localhost:4040/api/v1/applications/%s/stages", sparkAppID)
+	response, err = getResponseFromSparkMonitoringSvc(url)
+	if err != nil {
+		return "", fmt.Errorf("failed to get response from Spark Monitoring service at %s, %v", url, err)
+	}
+	var getStagesResult []map[string]interface{}
+	json.Unmarshal([]byte(response), &getStagesResult)
+	if len(getStagesResult) < 1 {
+		return "", fmt.Errorf("wrong number of Spark Application stages, expected more than 1, got %d", len(getStagesResult))
```
salv-orlando, line 122: According to condition you are expecting at least one stage, not more than one

salv-orlando, line 122: According to condition you are expecting at least one stage, not more than one

pkg/theiactl/commands/check.go
```diff
@@ -72,12 +78,88 @@ $ theiactl policyreco check --id e998433e-accb-4888-9fc8-06563f073e86
...
+	}
+	completedStages := 0
+	for _, stage := range getStagesResult {
+		if stage["status"] == "COMPLETE" || stage["status"] == "SKIPPED" {
+			completedStages++
+		}
+	}
+	return fmt.Sprintf(": %d/%d (%d%%) stages completed", completedStages, len(getStagesResult), completedStages*100/len(getStagesResult)), nil
```
salv-orlando, line 130: pedant comment: calculate `len(getStageResult)` once and then use value in the return statement above

salv-orlando, line 130: pedant comment: calculate `len(getStageResult)` once and then use value in the return statement above

pkg/theiactl/commands/check.go
```diff
@@ -72,12 +78,88 @@ $ theiactl policyreco check --id e998433e-accb-4888-9fc8-06563f073e86
...
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		return nil, getErr
+	}
+	if res.Body != nil {
+		defer res.Body.Close()
+	}
+	body, readErr := ioutil.ReadAll(res.Body)
```
salv-orlando, line 156: Here `res` should never be `nil`. But in that case should we instead return error?

salv-orlando, line 156: Here `res` should never be `nil`. But in that case should we instead return error?

pkg/theiactl/commands/result.go
```diff
@@ -49,9 +50,34 @@ $ theiactl policyreco result --id e998433e-accb-4888-9fc8-06563f073e86
 		if err != nil {
 			return fmt.Errorf("couldn't create k8s client using given kubeconfig, %v", err)
 		}
-		// TODO: setup port forwarder using clientset.
-		// TODO: get the recommendation result from ClickHouse using recoID
-		fmt.Printf("recoID: %v, clientset: %v\n", recoID, clientset)
+		err = CheckClickHousePod(clientset)
```
salv-orlando, line 53: I think the aim of this check is to prevalidate that we have a running clickhouse pod.
I do not think it adds much overhead, and it could be fine for the time being.

In the future I think we should have a more precise solution for validating the health status of the flow visibility solution

salv-orlando, line 53: I think the aim of this check is to prevalidate that we have a running clickhouse pod.
I do not think it adds much overhead, and it could be fine for the time being.

In the future I think we should have a more precise solution for validating the health status of the flow visibility solution

pkg/theiactl/commands/utils.go
```diff
@@ -48,10 +52,85 @@ func PolicyRecoPreCheck(clientset *kubernetes.Clientset) error {
...
 
+func CheckClickHousePod(clientset *kubernetes.Clientset) error {
+	// Check the ClickHouse deployment in flow-visibility namespace
+	pods, err := clientset.CoreV1().Pods(flowVisibilityNS).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "app=clickhouse",
+	})
+	if err != nil {
+		return fmt.Errorf("error %v when finding the ClickHouse Pod, please check the deployment of the ClickHouse", err)
+	}
+	if len(pods.Items) < 1 {
```
salv-orlando, line 67: Do you think it might make sense to verify pod status is running?

salv-orlando, line 67: Do you think it might make sense to verify pod status is running?

pkg/theiactl/commands/result.go
```diff
@@ -49,9 +50,34 @@ $ theiactl policyreco result --id e998433e-accb-4888-9fc8-06563f073e86
...
+		connect, err := ConnectClickHouse(clientset, databaseURL)
+		if err != nil {
+			return fmt.Errorf("error when connecting to ClickHouse, %v", err)
+		}
+		query := "SELECT yamls FROM recommendations WHERE id = (?);"
+		err = connect.QueryRow(query, recoID).Scan(&recoResult)
+		if err != nil {
+			return fmt.Errorf("get recommendation result failed of recommendation job with id %s: %v", recoID, err)
+		}
+		fmt.Print(recoResult)
```
salv-orlando, line 80: I assume more work might be needed on this PR for output formatting? cc @dreamtalen

salv-orlando, line 80: I assume more work might be needed on this PR for output formatting? cc @dreamtalen

salv-orlando (commented):
More comments about check and result commands

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+- [Command-line Usage](#command-line-usage)
+  - [Start a policy recommendation job](#start-a-policy-recommendation-job)
+  - [Check the status of a policy recommendation job](#check-the-status-of-a-policy-recommendation-job)
+  - [Get the result of a policy recommendation job](#get-the-result-of-a-policy-recommendation-job)
+<!-- /toc -->
+
+## Introduction
+
+Network policy recommendation is a feature that utilizes the flow records collected by
+[Grafana Flow Collector](network-flow-visibility.md#grafana-flow-collector) and generate
```
annakhm, line 17: generate ->generates

annakhm, line 17: generate ->generates

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+Kubernetes or [Antrea network policies](https://github.com/antrea-io/antrea/blob/main/docs/antrea-network-policy.md).
```
annakhm, line 18: Since `Antrea Network Policies` is a link, I'd suggest to change `Kubernetes` to `Kubernetes Network Policies` for clarity (perhaps also with a link?)

annakhm, line 18: Since `Antrea Network Policies` is a link, I'd suggest to change `Kubernetes` to `Kubernetes Network Policies` for clarity (perhaps also with a link?)

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+```bash
+$ kubectl get pods -n flow-aggregator
+NAME                                          READY   STATUS    RESTARTS   AGE
+policy-reco-spark-operator-56c4cb454c-4vhfh   1/1     Running   0          2m19s
+```
+
+## Command-line Usage
+
+Currently, the network policy recommendation feature only supports command line interaction through Theiactl.
+Theiactl is the Theia command-line tool. To get more information about Theiactl, please refer to this [doc](./theiactl.md). We have 3 Theiactl commands for the policy recommendation feature:
```
annakhm, line 60: Following `antctl` docs convention, I think we should use lowercase for `theiactl` unless beginning of sentence?
https://github.com/antrea-io/antrea/blob/main/docs/antctl.md

annakhm, line 60: Following `antctl` docs convention, I think we should use lowercase for `theiactl` unless beginning of sentence?
https://github.com/antrea-io/antrea/blob/main/docs/antctl.md

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+
+```bash
+$ theiactl policyreco start
+Policy recommendation start successfully, id is e998433e-accb-4888-9fc8-06563f073e86
+```
+
+### Check the status of a policy recommendation job
+
+After we start a policy recommendation job, we could use the `theiactl policyreco check`
+command can check the current status of that job.
```
annakhm, line 81: can -> to

annakhm, line 81: can -> to

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+
+For the policy recommendation job we just start above, we could check the status of it by:
```
annakhm, line 83: I think this line is redundant (rephrases the line above)?

annakhm, line 83: I think this line is redundant (rephrases the line above)?

pkg/theiactl/commands/utils.go
```diff
@@ -0,0 +1,136 @@
...
+		return fmt.Errorf("can't find the ClickHouse Pod, please check the deployment of ClickHouse")
+	}
+	return nil
+}
+
+func getServicePort(clientset *kubernetes.Clientset, name string) (int32, error) {
+	var servicePort int32
+	service, err := clientset.CoreV1().Services(flowVisibilityNS).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return servicePort, fmt.Errorf("error %v when finding Service %s, please check the deployment of the ClickHouse", err, name)
```
annakhm, line 77: If this function would be used for other services, we shouldn't mention `Clickhouse` here? Otherwise name should reflect that this is Clickhouse-specific.

pkg/theiactl/commands/utils.go
```diff
@@ -0,0 +1,136 @@
...
+	if servicePort == 0 {
+		return servicePort, fmt.Errorf("error %v when finding Service %s, please check the deployment of the ClickHouse", err, name)
+	}
+	return servicePort, nil
+}
+
+// Connect to ClickHouse in a loop
+func ConnectClickHouse(clientset *kubernetes.Clientset, url string) (*sql.DB, error) {
+	// Get the ClickHouse connection secret
+	secret, err := clientset.CoreV1().Secrets("flow-visibility").Get(context.TODO(), "clickhouse-secret", metav1.GetOptions{})
```
annakhm, line 93: "flow-visibility"->flowVisibilityNS

pkg/theiactl/commands/root.go
yanjunz97: I think `main.main()` here refers to the main function in `theiactl.go` as it is in package main.

Makefile
yanjunz97: Added

pkg/theiactl/commands/check.go
yanjunz97: Updated the error message.

pkg/theiactl/commands/check.go
yanjunz97: id is different from SparkAppId. But it's true that we do not need this parameter in this function any more. Removed it now, thanks!

yanjunz97: I notice there are some duplicated comments, just close them.

pkg/theiactl/commands/check.go
yanjunz97: Updated.

pkg/theiactl/commands/check.go
yanjunz97: Updated.

pkg/theiactl/commands/check.go
yanjunz97: Added the check.

pkg/theiactl/commands/utils.go
yanjunz97: Added.

pkg/theiactl/commands/result.go
yanjunz97: As the result in stored in YAML format in the database, we may not need to add output formatting here. User can save the output as YAML file and apply it with the current output.

pkg/theiactl/commands/utils.go
yanjunz97: As currently no other service requires this function, rename it to Clickhouse-specific.

pkg/theiactl/commands/utils.go
yanjunz97: Updated, thanks!

yanjunz97:
I notice that `theiactl recopolicy start` starts the recommendation job but does not clean it, which means the the policy-reco driver pod remained in Completed status when the job finished. Is that possible to delete the pod and related svc automatically somewhere?

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+To see all options and usage examples of these commands, you may run `theiactl policyreco [subcommand] --help`. In the following sections, we will go through a simple example to utilize these commands with default options.
+
+### Start a policy recommendation job
+
+The `theiactl policyreco start` command can start a new policy recommendation job.
+If the new policy recommendation job is created successfully, the `recommendation ID` of this job will be returned:
+
+```bash
+$ theiactl policyreco start
+Policy recommendation start successfully, id is e998433e-accb-4888-9fc8-06563f073e86
```
heanlan, line 75: "A new policy recommendation job is created successfully, id is"

pkg/theiactl/commands/check.go
```diff
@@ -0,0 +1,176 @@
...
+
+	sparkv1 "antrea.io/theia/third_party/sparkoperator/v1beta2"
+)
+
+// checkCmd represents the check command
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "Check the status of a policy recommendation Spark job",
+	Long: `Check the current status of a policy recommendation Spark job by ID.
+It will return the status of this Spark application like SUBMITTED, RUNNING, COMPLETED, or FAILED.`,
```
ziyouw, line 39: Not sure what does "Spark application" mean, is it different with "Spark job"?

pkg/theiactl/commands/check.go
```diff
@@ -0,0 +1,176 @@
...
+			state += stateProgress
+		}
+		fmt.Printf("Status of this policy recommendation job is %s\n", state)
+		return nil
+	},
+}
+
+func getPolicyRecommendationProgress(baseUrl string) (string, error) {
+	// Get the id of current spark application
+	url := fmt.Sprintf("%s/api/v1/applications", baseUrl)
```
ziyouw, line 103: For a non-existent id (for a typo issue), what state we will show?

yanjunz97: For a non-existent id, we will fail when getting status from Spark Application CRD instance with some error like 
```
Error: sparkapplications.sparkoperator.k8s.io "<non-existent id>" not found
```

pkg/theiactl/commands/policyreco.go
```diff
@@ -0,0 +1,42 @@
...
+		fmt.Println("Error: must also specify a subcommand like start, check or result")
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(policyrecoCmd)
+	rootCmd.PersistentFlags().StringP(
+		"kubeconfig",
+		"k",
+		"~/.kube/config",
```
annakhm, line 39: I don't think alias path can work here (needs to be converted to absolute path)

dreamtalen:
> I notice that `theiactl recopolicy start` starts the recommendation job but does not clean it, which means the the policy-reco driver pod remained in Completed status when the job finished. Is that possible to delete the pod and related svc automatically somewhere?

Resolved with offline discussion:
It's possible to set a timeout and delete the completed driver pod automatically. I didn't do that because this will let the check command can't find previously finished jobs with ID on the k8s API server. However, we could let the check command visit the Clickhouse DB first to see if this ID has corresponding results stored in the DB.

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+Kubernetes or [Antrea network policies](https://github.com/antrea-io/antrea/blob/main/docs/antrea-network-policy.md).
+The goal of this feature is to aid the cluster admin or developers to secure
+their applications and control network traffic based on the idea of the Zero Trust model.
+
+## Deployment
+
+To use the network policy recommendation feature, please follow this [doc](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md) to deploy Flow Exporter and Flow Aggregator first.
+Please update the `#podLabels: false` to `podLabels: true` inside
+[Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1),
+because policy recommendation feature needs Pod lables information to recommend network policies.
```
salv-orlando, line 27: typo: lables/labels

docs/network-policy-recommendation.md
salv-orlando: Sorry to be pedant:

Successfully created(or started) policy recommendation job #######

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Check the status of a policy recommendation job
+
+After we start a policy recommendation job, we could use the `theiactl policyreco check`
+command can check the current status of that job.
+
+For the policy recommendation job we just start above, we could check the status of it by:
+
+```bash
+$ theiactl policyreco check --id e998433e-accb-4888-9fc8-06563f073e86
```
salv-orlando, line 86: usability question: does the cobra framework take positional argument so that a user could write:

`theiactl policyreco check e998433e-accb-4888-9fc8-06563f073e86`

pkg/theiactl/commands/result.go
salv-orlando: I was thinking about indentation, but no reason to make a concern about it.

pkg/theiactl/commands/check.go
```diff
@@ -0,0 +1,207 @@
...
+			service := fmt.Sprintf("policy-reco-%s-ui-svc", recoID)
+			if useClusterIP {
+				serviceIP, servicePort, err := GetServiceAddr(clientset, service)
+				if err != nil {
+					return fmt.Errorf("error when getting Spark Monitor Service address: %v", err)
+				}
+				endpoint = fmt.Sprintf("tcp://%s:%d", serviceIP, servicePort)
+			} else {
+				servicePort := 4040
+				listenAddress := "localhost"
```
annakhm, line 96: I think we wanted this to be k8s API IP so that theiactl could be run externally? (same in other places)

yanjunz97: I thought we were forwarding the service port to the machine which runs theiactl? What does K8s API IP refers to?

annakhm: AFAIU we need to expose the port on k8s server API (Host in k8s config)

Makefile
salv-orlando: thanks!

pkg/theiactl/commands/check.go
salv-orlando: I don't think kubectl port forward will open port on the K8S API server, will it?
That seems the wrong thing to do from a security perspective.

I seem to recall it's

localhost:localport <---> K8SAPISERVER <---> Target Pod:Target port

pkg/theiactl/commands/check.go
salv-orlando: thanks!

pkg/theiactl/commands/root.go
salv-orlando: That was just a minor and somewhat cheeky comment, let's ignore it!

pkg/theiactl/commands/check.go
annakhm: Yes, sorry, you're both right, API server acts as a proxy here.

yanjunz97: Thanks for the explanation!

docs/network-policy-recommendation.md
yanjunz97: @salv-orlando This support is added in PR #31

pkg/theiactl/commands/start.go
```diff
@@ -0,0 +1,378 @@
...
+		if err != nil {
+			return err
+		}
+		matchResult, err = regexp.MatchString(k8sQuantitiesReg, executorMemory)
+		if err != nil || !matchResult {
+			return fmt.Errorf("executor_memory should conform to the Kubernetes convention")
+		}
+		sparkResourceArgs.executorMemory = executorMemory
+
+		kubeconfig, err := cmd.Flags().GetString("kubeconfig")
```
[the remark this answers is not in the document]
dreamtalen: We have a default value for kubeconfig, so the `MarkFlagRequired` may become redundant.

pkg/theiactl/commands/check.go
dreamtalen: Spark application is the name of CRD defined by Spark Operator. Sorry to be confusing here, I will rephrase these comments.

pkg/theiactl/commands/root.go
dreamtalen: Thanks, changed.

pkg/theiactl/commands/start.go
dreamtalen: Make sense to me, will handle this after the naming is finalized.

pkg/theiactl/commands/start.go
dreamtalen: Sure, added

pkg/theiactl/commands/start.go
dreamtalen: Right, I changed to use dashes for the flags in CLI.

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+helm repo add spark-operator https://googlecloudplatform.github.io/spark-on-k8s-operator
+helm install policy-reco spark-operator/spark-operator --namespace flow-visibility --set image.tag=latest
+```
+
+This will install the Kubernetes Operator for Apache Spark into the `flow-visibility` namespace.
+
+Once we finish these deployment steps, we should see `policy-reco-spark-operator` Pods running inside the `flow-visibility` namespace:
+
+```bash
+$ kubectl get pods -n flow-aggregator
```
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed

docs/network-policy-recommendation.md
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed

docs/network-policy-recommendation.md
dreamtalen: Thanks, addressed

docs/network-policy-recommendation.md
dreamtalen: Thanks, addressed

docs/network-policy-recommendation.md
dreamtalen: Thanks, addressed

docs/network-policy-recommendation.md
dreamtalen: Thanks, addressed

pkg/theiactl/commands/policyreco.go
dreamtalen: Thanks, I will change to $KUBECONFIG to be consistent with antctl

antoninbas (commented):
Small comments:
* If we want to add a `--wait` option, the I think `run` makes more sense than `wait` as a command name. It is more consistent with other tools that I am aware of as well (e.g., sonobuoy).
* I feel like the CLI tool could be simply `theia` instead of `theiactl`

dreamtalen:
> Small comments:
> 
> * If we want to add a `--wait` option, the I think `run` makes more sense than `wait` as a command name. It is more consistent with other tools that I am aware of as well (e.g., sonobuoy).
> * I feel like the CLI tool could be simply `theia` instead of `theiactl`

Thank Antonin, `--wait` option has been added.

I will change the naming later if no more suggestions from others, the latest layout suggested by Antonin and Jianjun is:
```
theia policy-recommendation run
theia policy-recommendation check
theia policy-recommendation retrieve
```
Also adding `pr` as an alias of `policy-recommendation`:
```
theia pr run
theia pr check
theia pr retrieve
```

annakhm:
Perhaps `theia pr status` rather than `theia pr check`? Just a thought, no strong opinion :)

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,129 @@
...
+- [Deployment](#deployment)
+- [Command-line Usage](#command-line-usage)
+  - [Start a policy recommendation job](#start-a-policy-recommendation-job)
+  - [Check the status of a policy recommendation job](#check-the-status-of-a-policy-recommendation-job)
+  - [Get the result of a policy recommendation job](#get-the-result-of-a-policy-recommendation-job)
+<!-- /toc -->
+
+## Introduction
+
+Network policy recommendation is a feature that utilizes the flow records collected by
```
jianjuns, line 16: NetworkPolicy to follow our naming conventions.

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,129 @@
...
+[Grafana Flow Collector](network-flow-visibility.md#grafana-flow-collector) and generates
+[Kubernetes network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/)
+or [Antrea network policies](https://github.com/antrea-io/antrea/blob/main/docs/antrea-network-policy.md).
+The goal of this feature is to aid the cluster admin or developers to secure
+their applications and control network traffic based on the idea of the Zero Trust model.
```
jianjuns, line 21: It is not clear what does "control network" mean. Probably remove it.

docs/network-policy-recommendation.md
jianjuns, line 16: I feel better to talk about what policy recommendation does, before how.

So, we can say something like "Theia NetworkPolicy Recommendation recommends the NetworkPolicy configuration to secure Kubernetes network and applications". And then "It analyzes the network flows collected by Grafana Flow Collector to generate Kubernetes NetworkPolicies and Antrea NetworkPolicies".

Assume @salv-orlando could have some good ideas.

dreamtalen:
> Perhaps `theia pr status` rather than `theia pr check`? Just a thought, no strong opinion :)

Thank Anna, `status` sounds good to me too. 
I found `sonoboy` also has a `status` command: `Gets a summarized status of a sonobuoy run`

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Deployment
+
+To use the network policy recommendation feature, please follow this [doc](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md) to deploy Flow Exporter and Flow Aggregator first.
```
jianjuns, line 25: We received a valid feedback the deployment steps are too complex to follow, and can be a big challenge for a normal user. Could we have an all-in-one quick-start to deploy everything, rather than refer to multiple docs? At least, we should have a quick-start to deploy Grafana Flow Collector with default settings, and then this doc can point to that.

I am also thinking probably better to split the documentation part to a separate PR, and so we can update/review it in parallel with the code.

@salv-orlando

docs/network-policy-recommendation.md
jianjuns: Then we should just document that way (assuming it is simpler). Again, consider create a separate PR just for documentation.

docs/network-policy-recommendation.md
```diff
@@ -0,0 +1,129 @@
...
+
+Once we finish these deployment steps, we should see the `policy-reco-spark-operator` Pod running inside the `flow-visibility` namespace:
+
+```bash
+$ kubectl get pods -n flow-visibility
+NAME                                          READY   STATUS    RESTARTS   AGE
+policy-reco-spark-operator-56c4cb454c-4vhfh   1/1     Running   0          2m19s
+```
+
+## Command-line Usage
```
jianjuns, line 58: Name it: "Perform NetworkPolicy Recommendation" to be explicit?

docs/network-policy-recommendation.md
dreamtalen: Thank Jianjun for reviewing this. Sure, let's move to https://github.com/antrea-io/theia/pull/36 for further reviews of docs.

pkg/theiactl/commands/check.go
```diff
@@ -0,0 +1,210 @@
...
+		"id",
+		"i",
+		"",
+		"ID of the policy recommendation Spark job",
+	)
+	checkCmd.Flags().Bool(
+		"use-cluster-ip",
+		false,
+		`Enable this option will use ClusterIP instead of port forwarding when connecting to the Spark Monitoring Service.
+It can only be used when running in cluster.`,
```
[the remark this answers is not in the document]
yanjunz97: Updated, thanks!

pkg/theiactl/commands/check.go
```diff
@@ -0,0 +1,210 @@
...
+	checkCmd.Flags().StringP(
+		"id",
+		"i",
+		"",
+		"ID of the policy recommendation Spark job",
+	)
+	checkCmd.Flags().Bool(
+		"use-cluster-ip",
+		false,
+		`Enable this option will use ClusterIP instead of port forwarding when connecting to the Spark Monitoring Service.
```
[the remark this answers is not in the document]
yanjunz97: Updated.

pkg/theiactl/commands/root.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "theiactl",
+	Short: "theiactl is the command line tool for Theia",
+	Long: `theiactl is the command line tool for Theia which provides access 
+to some network flow visibility capabilities`,
```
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed.

docs/theiactl.md
```diff
@@ -0,0 +1,38 @@
+# Theia
+
+Theiactl is the command-line tool for Theia. At the moment, theiactl supports running out-of-cluster only.
```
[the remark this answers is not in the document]
dreamtalen: Removed since we add the in-cluster option.

docs/network-policy-recommendation.md
dreamtalen: Thanks, addressed.

docs/network-policy-recommendation.md
dreamtalen: Thanks, addressed.

docs/network-policy-recommendation.md
dreamtalen: Thanks, addressed.

annakhm (approved):

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	)
+	policyRecommendationRunCmd.Flags().StringP(
+		"option",
+		"o",
+		"anp-deny-applied",
+		`Option of network isolation preference in policy recommendation.
+Currently we support 3 options:
+anp-deny-applied: Recommending allow ANP/ACNP policies, with default deny rules only on applied to Pod labels which have allow rules recommended.
+anp-deny-all: Recommending allow ANP/ACNP policies, with default deny rules for whole cluster.
+k8s-np: Recommending allow K8s network policies, with no deny rules at all`,
```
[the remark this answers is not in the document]
jianjuns: I feel we should remove "with no deny rules at all". K8s NP does support deny rule.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	)
+	policyRecommendationRunCmd.Flags().Int32(
+		"executor-instances",
+		1,
+		"Specify the number of executors for the spark application. Example values include 1, 2, 8, etc.",
+	)
+	policyRecommendationRunCmd.Flags().String(
+		"driver-core-request",
+		"200m",
+		`Specify the cpu request for the driver Pod. Values conform to the Kubernetes convention.
```
jianjuns, line 430: CPU

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	policyRecommendationRunCmd.Flags().String(
+		"driver-memory",
+		"512M",
+		`Specify the memory request for the driver Pod. Values conform to the Kubernetes convention.
+Example values include 512M, 1G, 8G, etc.`,
+	)
+	policyRecommendationRunCmd.Flags().String(
+		"executor-core-request",
+		"200m",
+		`Specify the cpu request for the executor Pod. Values conform to the Kubernetes convention.
```
jianjuns, line 442: CPU

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	policyRecommendationRunCmd.Flags().String(
+		"executor-memory",
+		"512M",
+		`Specify the memory request for the executor Pod. Values conform to the Kubernetes convention.
+Example values include 512M, 1G, 8G, etc.`,
+	)
+	policyRecommendationRunCmd.Flags().Bool(
+		"wait",
+		false,
+		"Enable this option will hold and wait the whole policy recommendation job finished.",
```
jianjuns, line 454: finished -> finishes

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	policyRecommendationRunCmd.Flags().String(
+		"clickhouse-endpoint",
+		"",
+		"The ClickHouse Service endpoint. (Only works when wait is enabled)",
+	)
+	policyRecommendationRunCmd.Flags().Bool(
+		"use-cluster-ip",
+		false,
+		`Enable this option will use ClusterIP instead of port forwarding when connecting to the ClickHouse Service.
+It can only be used when running in cluster. (Only works when wait is enabled)`,
```
jianjuns, line 465: What does "(Only works when wait is enabled)" mean?

Should we check "wait" is enabled and say: "It can only be used when running in cluster and when wait is enabled"?

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	)
+	policyRecommendationRunCmd.Flags().StringP(
+		"file",
+		"f",
+		"",
+		"The file path where you want to save the results. (Only works when wait is enabled)",
```
jianjuns, line 471: Ditto

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	)
+	policyRecommendationRunCmd.Flags().Bool(
+		"wait",
+		false,
+		"Enable this option will hold and wait the whole policy recommendation job finished.",
+	)
+	policyRecommendationRunCmd.Flags().String(
+		"clickhouse-endpoint",
+		"",
+		"The ClickHouse Service endpoint. (Only works when wait is enabled)",
```
jianjuns, line 459: Ditto

pkg/theia/commands/policy_recommendation_run.go
jianjuns: You used "result" in other places.

pkg/theia/commands/policy_recommendation_status.go
```diff
@@ -0,0 +1,218 @@
...
+	Short: "Check the status of a policy recommendation Spark job",
+	Long: `Check the current status of a policy recommendation Spark job by ID.
+It will return the status of this Spark application like SUBMITTED, RUNNING, COMPLETED, or FAILED.`,
+	Args: cobra.RangeArgs(0, 1),
+	Example: `
+Check the current status of job with ID e998433e-accb-4888-9fc8-06563f073e86
+$ theia policy-recommendation status --id e998433e-accb-4888-9fc8-06563f073e86
+Or
+$ theia policy-recommendation status e998433e-accb-4888-9fc8-06563f073e86
+Use Cluster IP when checking the current status of job with ID e998433e-accb-4888-9fc8-06563f073e86
```
jianjuns, line 47: ClusterIP

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+			Namespace(flowVisibilityNS).
+			Resource("sparkapplications").
+			Body(recommendationApplication).
+			Do(context.TODO()).
+			Into(response)
+		if err != nil {
+			return err
+		}
+		if waitFlag {
+			err = wait.Poll(statusCheckPollInterval, statusCheckPollTimeout, func() (bool, error) {
```
salv-orlando, line 309: if the 60min timeout expires, I assume we might still be able to check job completion via check and result commands, is that correct?

In that case it might be worth - unless too complex - to return an error message like:

Spark job wait timeout of 60minutes expired. Job <id> still running. Please check completion status for job <id> via CLI.

pkg/theia/commands/policy_recommendation_retrieve.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/google/uuid"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+// policyRecommendationRetrieveCmd represents the policy-recommendation retrieve command
+var policyRecommendationRetrieveCmd = &cobra.Command{
+	Use:   "retrieve",
```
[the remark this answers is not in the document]
dreamtalen: Hi Jianjun, I find sonobuoy have `retrieve` command to get results and another `results` command to inspect results. I prefer we could keep `retrieve` here.

pkg/theia/commands/policy_recommendation_retrieve.go
```diff
@@ -0,0 +1,241 @@
...
+	Short: "Get the recommendation result of a policy recommendation Spark job",
+	Long: `Get the recommendation result of a policy recommendation Spark job by ID.
+It will return the recommended network policies described in yaml.`,
```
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	executorCoreRequest string
+	executorMemory      string
+}
+
+// policyRecommendationRunCmd represents the policy recommendation run command
+var policyRecommendationRunCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run a new policy recommendation Spark job",
+	Long: `Run a new policy recommendation Spark job.
+Network policies will be recommended based on the flow records sent by Flow Aggregator.
```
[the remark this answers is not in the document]
dreamtalen: Thanks, removed this sentence.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+Must finish the deployment of Theia first, please follow the steps in 
+https://github.com/antrea-io/theia/blob/main/docs/network-policy-recommendation.md`,
```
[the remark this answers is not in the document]
dreamtalen: Thanks, removed this sentence.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+	Example: `Run a policy recommendation spark job with default configuration
```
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+		0,
+		"The limit on the number of flow records read from the database. 0 means no limit.",
+	)
+	policyRecommendationRunCmd.Flags().StringP(
+		"option",
+		"o",
+		"anp-deny-applied",
+		`Option of network isolation preference in policy recommendation.
+Currently we support 3 options:
+anp-deny-applied: Recommending allow ANP/ACNP policies, with default deny rules only on applied to Pod labels which have allow rules recommended.
```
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed.

pkg/theia/commands/policy_recommendation_run.go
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+		"limit",
+		"l",
+		0,
+		"The limit on the number of flow records read from the database. 0 means no limit.",
+	)
+	policyRecommendationRunCmd.Flags().StringP(
+		"option",
+		"o",
+		"anp-deny-applied",
+		`Option of network isolation preference in policy recommendation.
```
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+var policyRecommendationRunCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run a new policy recommendation Spark job",
+	Long: `Run a new policy recommendation Spark job.
+Network policies will be recommended based on the flow records sent by Flow Aggregator.
+Must finish the deployment of Theia first, please follow the steps in 
+https://github.com/antrea-io/theia/blob/main/docs/network-policy-recommendation.md`,
+	Example: `Run a policy recommendation spark job with default configuration
+$ theia policy-recommendation run
+Run an initial policy recommendation spark job with network isolation option anp-deny-applied and limit on last 10k flow records
```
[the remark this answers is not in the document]
dreamtalen: Thanks, addressed.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+		"{initial|subsequent} Indicates this recommendation is an initial recommendion or a subsequent recommendation job.",
+	)
+	policyRecommendationRunCmd.Flags().IntP(
+		"limit",
+		"l",
+		0,
+		"The limit on the number of flow records read from the database. 0 means no limit.",
+	)
+	policyRecommendationRunCmd.Flags().StringP(
+		"option",
```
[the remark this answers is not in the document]
dreamtalen: I prefer the `option` for now as it's shorter. Maybe we could add a prefix if we have other kinds of option in the future?

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+		"e",
+		"",
+		`The end time of the flow records considered for the policy recommendation.
+Format is YYYY-MM-DD hh:mm:ss in UTC timezone. No limit of the end time of flow records by default.`,
+	)
+	policyRecommendationRunCmd.Flags().StringP(
+		"ns-allow-list",
+		"n",
+		"",
+		`List of default traffic allow namespaces.
```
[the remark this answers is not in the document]
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+If no namespaces provided, Traffic inside Antrea CNI related namespaces: ['kube-system', 'flow-aggregator',
```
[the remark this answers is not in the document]
dreamtalen: Thanks addressed.

pkg/theia/commands/utils.go
```diff
@@ -0,0 +1,158 @@
...
+func CheckSparkOperatorPod(clientset kubernetes.Interface) error {
+	// Check the deployment of Spark Operator in flow-visibility ns
+	pods, err := clientset.CoreV1().Pods(flowVisibilityNS).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/instance=policy-reco,app.kubernetes.io/name=spark-operator",
+	})
+	if err != nil {
+		return fmt.Errorf("error %v when finding the policy-reco-spark-operator Pod, please check the deployment of the Spark Operator", err)
+	}
+	if len(pods.Items) < 1 {
+		return fmt.Errorf("can't find the policy-reco-spark-operator Pod, please check the deployment of the Spark Operator")
```
yanjunz97, line 65: Also replace `policy-reco-` by `policy-recommendation`?

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+'flow-visibility'] will be allowed by default.`,
+	)
+	policyRecommendationRunCmd.Flags().Bool(
+		"rm-labels",
+		true,
+		`Enable this option will remove automatically generated Pod labels including 'pod-template-hash',
```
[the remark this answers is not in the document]
dreamtalen: Remove here means that these automatically generated Pod labels will be ignored during policy recommendation and removed in the recommended policies. For example, a Pod have {app:pod-a, pod-template-hash:xxxxx}, in the generated policies, the appliedTo Pod labels will only contain {app:pod-a}

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+'controller-revision-hash', 'pod-template-generation'.`,
+	)
+	policyRecommendationRunCmd.Flags().Bool(
+		"to-services",
+		true,
+		`Use the toServices feature in ANP and recommendation toServices rules for Pod-to-Service flows,
+only works when option is 1 or 2.`,
```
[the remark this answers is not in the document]
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_run.go
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_run.go
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_run.go
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_run.go
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_run.go
dreamtalen: Thanks, change to use `result`.

pkg/theia/commands/policy_recommendation_run.go
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_status.go
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		sparkResourceArgs.executorInstances = executorInstances
+
+		driverCoreRequest, err := cmd.Flags().GetString("driver-core-request")
+		if err != nil {
+			return err
+		}
+		matchResult, err := regexp.MatchString(k8sQuantitiesReg, driverCoreRequest)
+		if err != nil || !matchResult {
+			return fmt.Errorf("driver-core-request should conform to the Kubernetes convention")
```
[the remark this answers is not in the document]
dreamtalen: Thanks addressed.

pkg/theia/commands/policy_recommendation_retrieve.go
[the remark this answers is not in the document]
jianjuns: Sure. Let us if others have an opinion.

pkg/theia/commands/policy_recommendation_run.go
dreamtalen: Sure, added a condition for time out usecase.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,473 @@
...
+			err = wait.Poll(statusCheckPollInterval, statusCheckPollTimeout, func() (bool, error) {
+				state, err := getPolicyRecommendationStatus(clientset, recommendationID)
+				if err != nil {
+					return false, err
+				}
+				if state == "COMPLETED" {
+					return true, nil
+				}
+				if state == "FAILED" || state == "SUBMISSION_FAILED" || state == "FAILING" || state == "INVALIDATING" {
+					return false, fmt.Errorf("policy recommendation job failed, state: %s", state)
```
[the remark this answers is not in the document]
dreamtalen: `return false, error` will also break the wait.Poll()

pkg/theia/commands/policy_recommendation_run.go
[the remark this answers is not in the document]
jianjuns: The problem is all the flags with a value are called options. How about policy-type?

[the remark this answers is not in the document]
dreamtalen: Sounds good to me, I will change it to `policy-type`.

pkg/theia/commands/policy_recommendation_run.go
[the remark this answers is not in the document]
jianjuns: Then it should be "exclude"?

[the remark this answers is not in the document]
dreamtalen: Got it, will change it to "exclude"

Makefile
```diff
@@ -165,3 +166,18 @@ policy-recommendation:
 	docker tag antrea/theia-policy-recommendation:$(DOCKER_IMG_VERSION) antrea/theia-policy-recommendation
 	docker tag antrea/theia-policy-recommendation:$(DOCKER_IMG_VERSION) projects.registry.vmware.com/antrea/theia-policy-recommendation
 	docker tag antrea/theia-policy-recommendation:$(DOCKER_IMG_VERSION) projects.registry.vmware.com/antrea/theia-policy-recommendation:$(DOCKER_IMG_VERSION)
+
+.PHONY: theia
+theia:
+	@mkdir -p $(BINDIR)
+	GOOS=linux $(GO) build -o $(BINDIR) $(GOFLAGS) -ldflags '$(LDFLAGS)' antrea.io/theia/pkg/theia
+
+# Add the darwin version binary to help dev&test on Mac for now
```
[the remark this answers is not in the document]
dreamtalen: OK, I'll remove it now.

pkg/theia/commands/utils.go
dreamtalen: Sure, I will leave the Labels as it is for now before #31 is merged.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,476 @@
...
+
+		recommendationID := uuid.New().String()
+		recoJobArgs = append(recoJobArgs, "--id", recommendationID)
+		recommendationApplication := &sparkv1.SparkApplication{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "sparkoperator.k8s.io/v1beta2",
+				Kind:       "SparkApplication",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "policy-recommendation-" + recommendationID,
```
yanjunz97, line 242: We may need a shorter name here as the UI service seems to fail to start if the name is too long. Find the following errors in the spark operator log:
```
I0604 01:07:15.780425      10 controller.go:263] Starting processing key: "flow-visibility/policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8"
I0604 01:07:15.780611      10 sparkui.go:282] Creating a service policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8-ui-svc for the Spark UI for application policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8
I0604 01:07:15.780861      10 event.go:282] Event(v1.ObjectReference{Kind:"SparkApplication", Namespace:"flow-visibility", Name:"policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8", UID:"9a99144f-3ef3-4ee2-bc81-57232f48cb2d", APIVersion:"sparkoperator.k8s.io/v1beta2", ResourceVersion:"18587", FieldPath:""}): type: 'Normal' reason: 'SparkApplicationAdded' SparkApplication policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8 was added, enqueuing it for submission
E0604 01:07:15.821772      10 controller.go:670] failed to create UI service for SparkApplication flow-visibility/policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8: Service "policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8-ui-svc" is invalid: metadata.name: Invalid value: "policy-recommendation-07fdd999-a7ab-4569-b918-8330237e32d8-ui-svc": must be no more than 63 characters
```

dreamtalen: Thanks Yanjun pointing out this. I propose we could change the job name to `pr-<id>`. I plan to make this change tomorrow if no objections.

salv-orlando: ops, thanks for spotting it @yanjunz97  - let's wait for a PR update from @dreamtalen

dreamtalen: Updated.

pkg/theia/commands/policy_recommendation_retrieve.go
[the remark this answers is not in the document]
salv-orlando: I do not have a strong opinion, both terms are good for me.

docs/network-policy-recommendation.md
salv-orlando: I believe we have achieved consensus on naming.
Resolving this comment.

pkg/theiactl/commands/start.go
salv-orlando: Thanks!

pkg/theiactl/commands/start.go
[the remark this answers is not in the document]
salv-orlando: sounds good, resolving conversaiton.

docs/network-policy-recommendation.md
salv-orlando: I added some comments on  doc PR #36.
@dreamtalen  should this commit be amended so that documentation is entirely contained in PR #36?

salv-orlando (approved):
I don't have any additional comment. PR looks good to me.

pkg/theia/commands/policy_recommendation_retrieve_test.go
```diff
@@ -0,0 +1,57 @@
...
+	recoID := "db2134ea-7169-46f8-b56d-d643d4751d1d"
+	expectedResult := "recommend-allow-acnp-kube-system-rpeal"
+	db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
+	assert.NoError(t, err)
+	defer db.Close()
+	resultRow := sqlmock.NewRows([]string{"yamls"}).AddRow(expectedResult)
+	mock.ExpectQuery("SELECT yamls FROM recommendations WHERE id = (?);").WithArgs(recoID).WillReturnRows(resultRow)
+	result, err := getResultFromClickHouse(db, recoID)
+	assert.NoError(t, err)
+	assert.Equal(t, expectedResult, result)
```
ziyouw, line 56: For UT, only covering happy path may be not enough, we may also consider some negative cases later.

ziyouw (approved):
For UT, we can consider to cover more cases later.

pkg/theia/commands/policy_recommendation_run.go
```diff
@@ -0,0 +1,476 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	sparkv1 "antrea.io/theia/third_party/sparkoperator/v1beta2"
+)
+
+const (
+	flowVisibilityNS        = "flow-visibility"
+	k8sQuantitiesReg        = "^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$"
+	sparkImage              = "antrea/theia-policy-recommendation:latest"
```
heanlan, line 38: change to "projects.registry.vmware.com/antrea/theia-policy-recommendation:latest" before getting merged?

pkg/theia/commands/policy_recommendation_retrieve_test.go
dreamtalen: Added some failure cases in the second commit, we could add more cases after this PR merged.

PR #30, opened 2022-05-27
Implement k8s port forwarding tooling for theiactl

annakhm:
In order to access theia services externally, theiactl can
make use of k8s port forwarder functionality. This change
provides tooling to start/stop port forwarding channel for
eather Pod+Port or Service+Port. The code is loosely based on
corresponding code in kubectl.

Signed-off-by: Anna Khmelnitsky <akhmelnitsky@vmware.com>


pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,179 @@
...
+	pods, err := pf.Clientset.CoreV1().Pods(pf.Namespace).List(context.TODO(), listOptions)
+
+	if err != nil {
+		return pf, fmt.Errorf("Failed to read Pods for Service %s: %v", service, err)
+	}
+	if len(pods.Items) == 0 {
+		return pf, fmt.Errorf("No Pods found for Service %s: %v", service, err)
+	}
+
+	pod := pods.Items[0]
```
salv-orlando, line 106: An important aspect of kubectl port forward is that you always forward to a pod, is that correct?
I don't think you can forward to a service and let k8s kube proxy do the endpoint selection

salv-orlando (commented):
Anna, can you also provide an example of how to use the port forwarder in the CLI command?

pkg/theiactl/portforwarder/portforwarder.go
annakhm: The API only allows forwarding to a pod, however kubectl allows to forward to a service, and behind the scenes it would choose first pod for the service (this logic is copied in this PR)

pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,179 @@
...
+// This code is based upon kubectl port-forward implementation
+func getContainerPortByServicePort(svc *v1.Service, port int) (int, error) {
+	for _, portspec := range svc.Spec.Ports {
+		if int(portspec.Port) != port {
+			continue
+		}
+		if svc.Spec.ClusterIP == v1.ClusterIPNone {
+			return port, nil
+		}
+		if portspec.TargetPort.Type == intstr.Int {
```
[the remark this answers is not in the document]
annakhm: I'm not sure how generic we want to make this support.

pkg/theiactl/portforwarder/portforwarder.go
salv-orlando: Understood, thanks for the explanation

pkg/theiactl/portforwarder/portforwarder.go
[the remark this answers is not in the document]
salv-orlando: It could be a todo action. So far our services are not using named ports, so not a must have.

annakhm:
Added comments in code to clarify the usage.

pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,185 @@
...
+	TargetPort    int
+	ListenAddress string
+	ListenPort    int
+	stopCh        chan struct{}
+}
+
+// This function creates Port Forwarder for a Pod
+// After creating Port Forwarder object, call Start() on it to start forwarding
+// channel and Stop() to terminate it
+func NewPortForwarder(config *rest.Config, namespace string, pod string, listenAddress string, listenPort int, targetPort int) (*PortForwarder, error) {
```
yanjunz97, line 47: Could we change the parameter order to something like "pod string, targetPort int, listenAddress string, listenPort int"? The current order seems a little bit strange.

pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,185 @@
...
+
+	return pf, nil
+}
+
+// This function creates Port Forwarder for a Service by finding first Pod
+// that belongs to the Service, and target Port for requested Service Port.
+// This code is based upon kubectl port-forward implementation
+// After creating Port Forwarder object, call Start() on it to start forwarding
+// channel and Stop() to terminate it
+func NewServicePortForwarder(config *rest.Config, namespace string, service string, listenAddress string, listenPort int, servicePort int) (*PortForwarder, error) {
```
yanjunz97, line 72: Same for the parameter order

pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,185 @@
...
+
+	var err error
+	pf.Clientset, err = kubernetes.NewForConfig(pf.Config)
+	if err != nil {
+		return pf, fmt.Errorf("Could not create kubernetes client: %v", err)
+	}
+
+	serviceObj, err := pf.Clientset.CoreV1().Services(pf.Namespace).Get(context.TODO(), service, metav1.GetOptions{})
+	if err != nil {
+		return pf, fmt.Errorf("Failed to read Service %s: %v", service, err)
```
yanjunz97, line 88: I think we do not capitalize the first letter of the message for `fmt.Errorf`

pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,185 @@
...
+// This code is based upon kubectl port-forward implementation
+// After creating Port Forwarder object, call Start() on it to start forwarding
+// channel and Stop() to terminate it
+func NewServicePortForwarder(config *rest.Config, namespace string, service string, listenAddress string, listenPort int, servicePort int) (*PortForwarder, error) {
+
+	pf := &PortForwarder{
+		Namespace:     namespace,
+		ListenAddress: listenAddress,
+		ListenPort:    listenPort,
+	}
```
yanjunz97, line 78: Miss `Config: config`?

pkg/theiactl/portforwarder/portforwarder.go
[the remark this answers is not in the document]
annakhm: Done, thanks

pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,185 @@
...
+	ListenAddress string
+	ListenPort    int
+	stopCh        chan struct{}
+}
+
+// This function creates Port Forwarder for a Pod
+// After creating Port Forwarder object, call Start() on it to start forwarding
+// channel and Stop() to terminate it
+func NewPortForwarder(config *rest.Config, namespace string, pod string, listenAddress string, listenPort int, targetPort int) (*PortForwarder, error) {
+
```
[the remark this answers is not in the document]
annakhm: Removed

pkg/theiactl/portforwarder/portforwarder.go
annakhm: I was following the order for `kubectl port-forward` command, but your suggestion makes more sense to me too, so I changed the order. Thanks!

pkg/theiactl/portforwarder/portforwarder.go
annakhm: Good catch, thanks!

pkg/theiactl/portforwarder/portforwarder.go
```diff
@@ -0,0 +1,189 @@
...
+		return pf, fmt.Errorf("failed to read Service %s: %v", service, err)
+	}
+
+	// find container port that corresponds to requested service port
+	pf.targetPort, err = getContainerPortByServicePort(serviceObj, servicePort)
+	if err != nil {
+		return pf, err
+	}
+
+	klog.Infof("Port forwarder requested for service %s/%s: %s:%d -> %d", namespace, service, listenAddress, servicePort, pf.targetPort)
```
yanjunz97, line 99: I understand the advantage of logs. But as theiactl uses this package to do port forwarding and displays status of spark job or yaml format results. It seems a little bit strange to have the log before the results, and this exposes unnecessary details for CTL users. Would it be better to only have this message when there is an error, or in a debug mode?

annakhm: Agreed, I would only expose this in verbose mode. I don't think initial theiactl PR includes verbose mode, but once its added we can involve it here.

PR #27, opened 2022-05-26
[WIP]Add e2e test for policy recommendation

heanlan:

PR #35, opened 2022-06-02
[e2e] Adjustments for recent K8S versions

salv-orlando:
- Retrieve cluster version before fetching first node
  to ensure control plane node is always correctly
  identified
- Tolerate both "master" and "control-plane" taints
  in pod specs

Signed-off-by: Salvatore Orlando <sorlando@vmware.com>


heanlan (approved):
There's some changes needed  in test-vmc.sh in order to pass the Jenkins test, which I've already included in #27 

We can safely merge this one.

PR #42, opened 2022-06-08
Fix FlowAggregator e2e test ANP deployment issue

heanlan:
The original e2e test cases only check IngressRuleTable/EgressTable flow
count, and use the result to decide the if the NP rule is realized. But
these two tables are used to realize K8s NP rules. For a normal ANP
rule, AntreaPolicyIngressRuleTable/AntreaPolicyEgressRuleTable are used,
but never checked in the case. Besides, the original case always uses
number 2 to check the rule flow existence, but the fact is there always
no less than 2 flows in IngressRule/EgressRule table, which are used to
bypass the established and related packets in a valid connection. So the
case actually doesn't check if the rule is realized or not.

Signed-off-by: wenyingd <wenyingd@vmware.com>
Co-authored-by: heanlan <hanlan@vmware.com>

### DIFF
--- a/test/e2e/flowaggregator_test.go
+++ b/test/e2e/flowaggregator_test.go
@@ -33,6 +33,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"antrea.io/antrea/pkg/agent/openflow"
 	"antrea.io/antrea/pkg/antctl"
 	"antrea.io/antrea/pkg/antctl/runtime"
 	secv1alpha1 "antrea.io/antrea/pkg/apis/crd/v1alpha1"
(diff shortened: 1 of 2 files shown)


salv-orlando (approved):
LGTM

test/e2e/flowaggregator_test.go
```diff
@@ -878,14 +883,17 @@ func deployK8sNetworkPolicies(t *testing.T, data *TestData, srcPod, dstPod strin
...
+	}
+	if err := data.WaitNetworkPolicyRealize(controlPlaneNodeName(), openflow.IngressRuleTable, egressTableInitFlowCount+1); err != nil {
+		t.Errorf("Error when waiting for egress Network Policy to be realized: %v", err)
 	}
 	t.Log("Network Policies are realized.")
 	return np1, np2
 }
 
-func deployAntreaNetworkPolicies(t *testing.T, data *TestData, srcPod, dstPod string) (anp1 *secv1alpha1.NetworkPolicy, anp2 *secv1alpha1.NetworkPolicy) {
+func deployAntreaNetworkPolicies(t *testing.T, data *TestData, srcPod, dstPod string, srcNode, dstNode string) (anp1 *secv1alpha1.NetworkPolicy, anp2 *secv1alpha1.NetworkPolicy) {
```
dreamtalen, line 896: nit: `srcPod, dstPod string, srcNode, dstNode string` -> `srcPod, dstPod, srcNode, dstNode string`

dreamtalen (approved):
a nit only, otherwise LGTM

PR #44, opened 2022-06-09
Add unit tests and Python workflow for Policy Recommendation job

### LINKED ISSUE #47: Add unit tests for policy recommendation job
Currently, policy recommendation Spark jobs are written in Python and lack unit tests covered. Since Python is an interpreted language, unit test coverage is important to ensure the correct functionalities. 

We need to add unit tests for these codes and Python CI on Github to automatically run unit tests when Python code changes are detected.

dreamtalen:
Fix #47 

Add UT and python workflow for policy recommendation spark code to improve reliability.

### DIFF
--- a/plugins/policy-recommendation/policy_recommendation_job_test.py
+++ b/plugins/policy-recommendation/policy_recommendation_job_test.py
@@ -30,21 +30,21 @@ def spark_session(request):
  
 @pytest.mark.parametrize("test_input, expected_flow_type", [
     (
-        ("", ""), 
+        (3, "", ""), 
         "pod_to_external"
     ),
     (
-        ("", '{"podname":"perftest-c"}'), 
+        (1, "", '{"podname":"perftest-c"}'), 
         "pod_to_pod"
     ),
     (
-        ("antrea-e2e/perftestsvc:5201", '{"podname":"perftest-c"}'), 
+        (1, "antrea-e2e/perftestsvc:5201", '{"podname":"perftest-c"}'), 
         "pod_to_svc"
     ),
 ])
 def test_get_flow_type(test_input, expected_flow_type):
-    destinationServicePortName, destinationPodLabels = test_input
-    flow_type = get_flow_type(destinationServicePortName, destinationPodLabels)
+    flowType, destinationServicePortName, destinationPodLabels = test_input
+    flow_type = get_flow_type(flowType, destinationServicePortName, destinationPodLabels)
     assert flow_type == expected_flow_type
 
 @pytest.mark.parametrize("test_input, expected_labels", [
@@ -113,7 +113,6 @@ table_name = "default.flows"
 def test_generate_sql_query(test_input, expected_sql_query):
     limit, start_time, end_time, unprotected = test_input
     sql_query = generate_sql_query(table_name, limit, start_time, end_time, unprotected)
-    print(expected_sql_query)
     assert sql_query == expected_sql_query
 
 @pytest.mark.parametrize("test_input, expected_policies", [
@@ -709,7 +708,7 @@ def test_generate_anp_ingress_rule(test_input, expected_ingress_rule):
 ])
 def test_recommend_k8s_policies(spark_session, flows_input, expected_policies):
     test_df = spark_session.createDataFrame(
-        flows_input, FLOW_TABLE_COLUMNS + ["flowType"]
+        flows_input, FLOW_TABLE_COLUMNS
     )
     recommend_k8s_polices_yamls = recommend_k8s_policies(test_df)
     recommend_k8s_polices_dicts = [
@@ -1165,7 +1164,7 @@ def test_recommend_k8s_policies(spark_session, flows_input, expected_policies):
 def test_recommend_antrea_policies(spark_session, test_input, expected_policies):
     flows_input, option, deny_rules, to_services = test_input
     test_df = spark_session.createDataFrame(
-        flows_input, FLOW_TABLE_COLUMNS + ["flowType"]
+        flows_input, FLOW_TABLE_COLUMNS
     )
     recommend_polices_yamls = recommend_antrea_policies(test_df, option, deny_rules, to_services)
     recommend_polices_dicts = [
--- a/plugins/policy-recommendation/policy_recommendation_job_test.py
+++ b/plugins/policy-recommendation/policy_recommendation_job_test.py
@@ -1,0 +1,1215 @@
+# Copyright 2022 Antrea Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


annakhm:
Do we want to consider using [unittest ](https://docs.python.org/3/library/unittest.html) package for the tests?

plugins/policy-recommendation/policy_recommendation_job_test.py
```diff
@@ -0,0 +1,1215 @@
...
+            and policy["metadata"]["name"].startswith("recommend-k8s-np-") == True
+        )
+        policy["metadata"]["name"] = expect_policy["metadata"]["name"]
+        assert policy == expect_policy
+
+
+@pytest.mark.parametrize("test_input, expected_policies", [
+    (
+        (flows_input, 1, True, False),
+        {
```
ziyouw, line 740: Should we put flow data and Json objects into other input/output files? Latter, if we update the NP logic and change the input/result, maybe it would be simple to update the files directly and keep the code unchaned.

dreamtalen: Hi Ziyou, yes we could definitely do that, just wondering whether it's a best practice or not.
If we choose to put test input and expected output into files, there will be a lot of .txt and .json files (at least two files for each test case). Is this our convention in other Python projects?

ziyouw: For test purpose, I think it is fine and it is also possiable for us to reuse the same test code for multi test cases.

ziyouw: For test purpose, I think it is fine and it is also possiable for us to reuse the same test code for multi test cases.

plugins/policy-recommendation/policy_recommendation_job_test.py
```diff
@@ -0,0 +1,1215 @@
...
+                    protocol = "TCP"
+                )
+            ]
+        )
+    ),
+])
+def test_generate_k8s_egress_rule(test_input, expected_egress_rule):
+    if test_input == "wrongformat":
+        with pytest.raises(SystemExit) as pytest_wrapped_e:
+                generate_k8s_egress_rule(test_input)
```
salv-orlando, line 411: can this call raise a SystemExit error?

plugins/policy-recommendation/policy_recommendation_job_test.py
```diff
@@ -0,0 +1,1215 @@
...
+        assert ingress_rule == expected_ingress_rule
+
+@pytest.mark.parametrize("test_input, expected_policy_name", [
+    (
+        "recommend-k8s-np",
+        "recommend-k8s-np-y0cq6"
+    ),
+    (
+        "recommend-allow-anp",
+        "recommend-allow-anp-y0cq6"
```
salv-orlando, line 466: I admit I don't understand how can guarantee a specific suffix is appended.

plugins/policy-recommendation/policy_recommendation_job_test.py
salv-orlando: These are dicts; while in most cases they can be stored as JSON, there are cases when they contain obect specs which might not be serializable.
The suggestion is valid, but maybe not to be scoped in this PR.

plugins/policy-recommendation/policy_recommendation_job_test.py
```diff
@@ -0,0 +1,1215 @@
...
+                },
+            },
+        },
+    ),
+])
+def test_recommend_antrea_policies(spark_session, test_input, expected_policies):
+    flows_input, option, deny_rules, to_services = test_input
+    test_df = spark_session.createDataFrame(
+        flows_input, FLOW_TABLE_COLUMNS + ["flowType"]
+    )
```
salv-orlando, line 1169: forgive my ignorance, is this creating a fake dataframe with data from `flows_input`?

salv-orlando (commented):
Overall looks good to me.
Did you already run coverage results locally or using codecov?

plugins/policy-recommendation/policy_recommendation_job_test.py
```diff
@@ -0,0 +1,1215 @@
...
+    ),
+    (
+        (100, "2022-01-01 00:00:00", "2022-01-01 23:59:59", True), 
+        "SELECT {} FROM {} WHERE ingressNetworkPolicyName == '' AND egressNetworkPolicyName == '' AND flowStartSeconds >= '2022-01-01 00:00:00' AND flowEndSeconds < '2022-01-01 23:59:59' GROUP BY {} LIMIT 100".format(", ".join(FLOW_TABLE_COLUMNS), table_name, ", ".join(FLOW_TABLE_COLUMNS))
+    ),
+])
+def test_generate_sql_query(test_input, expected_sql_query):
+    limit, start_time, end_time, unprotected = test_input
+    sql_query = generate_sql_query(table_name, limit, start_time, end_time, unprotected)
+    print(expected_sql_query)
```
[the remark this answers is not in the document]
dreamtalen: Good catch! deleted.

plugins/policy-recommendation/policy_recommendation_job_test.py
dreamtalen: By calling `random.seed(0)`, we could make sure the randomly generated suffix are the same.

plugins/policy-recommendation/policy_recommendation_job_test.py
dreamtalen: Yes, `flows_input` will be the data, and `FLOW_TABLE_COLUMNS + ["flowType"]` will be the schema of this fake dataframe for testing.

dreamtalen:
> Overall looks good to me. Did you already run coverage results locally or using codecov?

Yes, you could check latest coverage results on codecov: https://app.codecov.io/gh/antrea-io/theia/pulls

plugins/policy-recommendation/policy_recommendation_job_test.py
salv-orlando: Got it - so you're relying on seeding the number generator in the same way so that it always generates the same sequence. That should work as long as we run the tests on linux/amd64, as far as I recall

plugins/policy-recommendation/policy_recommendation_job_test.py
salv-orlando: Understood!

One thing we might consider for the polcy reco job is instead of having sys.exit throughout the code is to raise specific exception (we can also define exception types if needed) and then let those exception bubble up to the main function where we can use the exception type to exit with the appropriate status code.

This is more for code readability/maintanability, but will also allow us to capture precise exception types in unit tests.

We can iterate over this with another PR, for this PR it's perfectly fine to check for SystemExit

salv-orlando (approved):
PR LGTM. We can iterate with more PRs for additional coverage and other improvements.

PR #46, opened 2022-06-13
Export Pod logs when TestMain fails at deploying Antrea

heanlan:
This commit adds the Pods log exporting process to TestMain. The
major motivation is to help triage when the Antrea controller or
Antrea agent Pods fail to start.

### DIFF
--- a/test/e2e/fixture.go
+++ b/test/e2e/fixture.go
@@ -43,7 +43,7 @@ func ensureAntreaRunning(data *TestData) error {
 }
 
 func teardownTest(tb testing.TB, data *TestData) {
-	exportLogs(tb, data, "beforeTeardown", true)
+	exportLogs(tb, data, "beforeTeardown", true, false)
 	if empty, _ := IsDirEmpty(data.logsDirForTestCase); empty {
 		_ = os.Remove(data.logsDirForTestCase)
 	}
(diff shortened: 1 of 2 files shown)


test/e2e/fixture.go
```diff
@@ -254,6 +254,116 @@ func exportLogs(tb testing.TB, data *TestData, logsSubDir string, writeNodeLogs
...
+		logFile := filepath.Join(logsDir, fmt.Sprintf("%s-%s", nodeName, suffix))
+		f, err := os.Create(logFile)
+		if err != nil {
+			log.Printf("Error when creating log file '%s': '%v'", logFile, err)
+			return nil
+		}
+		return f
+	}
+	// export kubelet logs with journalctl for each Node. If the Nodes do not use journalctl we
+	// print a log message. If kubelet is not run with systemd, the log file will be empty.
```
yuntanghsu, line 341: Should it be " does not" or "running"?

heanlan: changed to "running", thanks

test/e2e/fixture.go
```diff
@@ -254,6 +254,116 @@ func exportLogs(tb testing.TB, data *TestData, logsSubDir string, writeNodeLogs
 	}
 }
 
+func exportMainTestLogs(data *TestData, logsSubDir string, writeNodeLogs bool) {
+	const timeFormat = "Jan02-15-04-05"
```
dreamtalen, line 258: There is a const `timeFormat` in function `exportLogs()`, maybe you could make that one accessible and reuse it.

test/e2e/fixture.go
```diff
@@ -254,6 +254,116 @@ func exportLogs(tb testing.TB, data *TestData, logsSubDir string, writeNodeLogs
...
+	err := createDirectory(logsDir)
+	if err != nil {
+		log.Printf("Error when creating logs directory '%s': %v", logsDir, err)
+		return
+	}
+	log.Printf("Exporting test logs to '%s'", logsDir)
+
+	// getPodWriter creates the file with name nodeName-podName-suffix. It returns nil if the
+	// file cannot be created. File must be closed by the caller.
+	getPodWriter := func(nodeName, podName, suffix string) *os.File {
```
dreamtalen, line 270: Oh actually all lines are duplicated from function `exportLogs()`, probably better to add an option for function `exportLogs()` rather than creating a new function?

heanlan: Yes, they are duplicated. The tricky thing is, `exportLogs` has an argument `tb testing.TB`, and TestMain doesn't have this interface. So for all the occurrence of `tb` in the function, we'll need to add a if statement. e.g. `tb.Logf` -> `log.Printf`. Do you have any suggestions?

dreamtalen: Thanks, looks much better now.

test/e2e/fixture.go
```diff
@@ -124,24 +124,26 @@ func forAllNodes(fn func(nodeName string) error) error {
...
 
-func exportLogs(tb testing.TB, data *TestData, logsSubDir string, writeNodeLogs bool) {
-	if tb.Skipped() {
-		return
-	}
-	// if test was successful and --logs-export-on-success was not provided, we do not export
-	// any logs.
-	if !tb.Failed() && !testOptions.logsExportOnSuccess {
-		return
+func exportLogs(tb *testing.TB, data *TestData, logsSubDir string, writeNodeLogs bool, testMain bool) {
```
dreamtalen, line 127: I feel a little weird about changing the parameter from `testing.TB` to `*testing.TB` but probably not a big deal for now.
Other changes LGTM.

heanlan: if we don't change testing.TB to a pointer, I don't know what parameter to pass here when making call from testMain.

dreamtalen: `testing.TB` is the interface common to T and B, so you could try passing an empty T struct.

heanlan: so true! updated. thanks Yongming.

dreamtalen (approved):
LGTM

PR #55, opened 2022-06-21
Support ClickHouse Cluster deployment

### LINKED ISSUE #53: Support ClickHouse cluster deployment
**Describe what you are trying to solve**
 Currently we only support the single node ClickHouse server, which lacks reliability and the ability to scale horizontally.

**Describe the solution you have in mind** 
Support ClickHouse cluster deployment with shards for horizontal scaling and replicas for reliability.

**Describe how your solution impacts user flows** 
ClickHouse cluster can manage flow records in a larger network with reliability for users.

**Describe the main design/architecture of your solution** 
Refer to the [ClickHouse Operator docs](https://github.com/Altinity/clickhouse-operator/blob/master/docs/replication_setup.md ) to enable the ClickHouse cluster deployment. Add corresponding values in Theia Helm Chart to make the deployment user-friendly.

yanjunz97:
Fix #53 

This PR adds support for ClickHouse cluster deployment with shards and replicas.


yuntanghsu:
Do we need to add podDistribution for clickhouse pod? Like ClickHouseAntiAffinity/ShardAntiAffinity/ReplicaAntiAffinity?

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,55 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
-[Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1).
+The ClickHouse TCP service is used by the Flow Aggregator. If you have changed
+the TCP port, please update the `clickHouse.databaseURL` in the Flow Aggregator
+Helm Chart values.
 
-The ClickHouse credentials are used in the Flow Aggregator. Them are also specified
-in `flow-aggregator.yml` as a Secret named `clickhouse-secret` as shown below.
-Please make the corresponding changes if you change the ClickHouse credentials.
+The ClickHouse credentials are used in the Flow Aggregator. If you have changed
+the ClickHouse credentials, please make the `clickHouse.connectionSecret` in the
```
dreamtalen, line 211: make -> update

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,55 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+
+The default affinity allows only one ClickHouse instance per Node. To change the
+affinity, please set `clickhouse.cluster.podDistribution` per your requirement.
+
+We recommend using Persistent Volumes if you are going to use ClickHouse cluster
+in production. ClickHouse cluster with in memory deployment is mainly for test
+purpose, as when a pod accidentally restarts, the memory will be cleared, which
+will lead to a complete data loss. In that case, please refer to the
+[ClickHouse doc](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication/#recovery-after-complete-data-loss)
+to recovery the data.
```
dreamtalen, line 246: recovery -> recover

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,55 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+[ClickHouse instructions](https://github.com/Altinity/clickhouse-operator/blob/master/docs/zookeeper_setup.md)
+on how to set up ZooKeeper and set `clickhouse.cluster.zookeeperHosts` to your
+ZooKeeper hosts.
+
+The default affinity allows only one ClickHouse instance per Node. To change the
+affinity, please set `clickhouse.cluster.podDistribution` per your requirement.
+
+We recommend using Persistent Volumes if you are going to use ClickHouse cluster
+in production. ClickHouse cluster with in memory deployment is mainly for test
+purpose, as when a pod accidentally restarts, the memory will be cleared, which
```
dreamtalen, line 243: pod -> Pod

docs/network-flow-visibility.md

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,55 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+To deploy ClickHouse cluster, please set `clickhouse.cluster.enable` to true and
+set `clickhouse.cluster.shards` and `clickhouse.cluster.replicas` per your
+requirement.
+
+ClickHouse uses [Apache ZooKeeper](https://zookeeper.apache.org/) for storing
+replicas meta information. By default, a ZooKeeper cluster using `emptyDir` with
+3 replicas will be deployed for ClickHouse cluster. Please make sure there are
+at least 3 nodes available in your Kubernetes cluster. To use a customized
+ZooKeeper cluster, please refer the
+[ClickHouse instructions](https://github.com/Altinity/clickhouse-operator/blob/master/docs/zookeeper_setup.md)
```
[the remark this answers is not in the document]
yanjunz97: Make sense! I update it to `ZooKeeper setup instructions for ClickHouse`. Please tell me if you have other suggestion, thanks!

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,54 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+
+ClickHouse uses [Apache ZooKeeper](https://zookeeper.apache.org/) for storing
+replicas meta information. By default, a ZooKeeper cluster using `emptyDir` with
+3 replicas will be deployed for ClickHouse cluster. Please make sure there are
+at least 3 Nodes available in your Kubernetes cluster. To use a customized
+ZooKeeper cluster, please refer to the
+[ZooKeeper setup instructions for ClickHouse](https://github.com/Altinity/clickhouse-operator/blob/master/docs/zookeeper_setup.md)
+and set `clickhouse.cluster.zookeeperHosts` to your ZooKeeper hosts.
+
+The default affinity allows only one ClickHouse instance per Node. To change the
```
yuntanghsu, line 237: Do we need to remind user to have 4 nodes setup?  Since the default shard=2, replicas=2 and we have podDistribution setting MaxNumberPerNode=1.
Or use ShardAntiAffinity, since we already have at least 3 Nodes?

yanjunz97: I think basically `MaxNumberPerNode=1` would be recommended to achieve reliability. But I'm worry about the redundant statement on different number of Nodes would be confusing. I tried to rephrase the words here. Please take a look whether it is clearer. Appreciate any suggestion!

yuntanghsu: I'm also thinking the number of nodes in zookeeper and clickchouse affinity will be confusing.
But I think current statement is clear enough for user to know where the numbers come from.

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
-Please refer to the [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
+Please refer to the Flow Aggregator Helm Chart [README](../build/charts/flow-aggregator/README.md)
+and [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
 to learn about more ClickHouse configuration options in Flow Aggregator.
 
+##### ClickHouse Cluster
+
+From Theia v0.2.0, we support deploying ClickHouse cluster with data replication.
+A ClickHouse cluster consists of shards. Shards refer to the servers that contain
+different parts of the data, which provides a horizontal cluster scaling strategy.
```
jianjuns, line 222: provides -> provide?

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+Each shard consists of one or more replica hosts to ensure reliability.
+
+To deploy ClickHouse cluster, please set `clickhouse.cluster.enable` to true and
```
jianjuns, line 225: "To deploy ClickHouse  as a cluster"?

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
-```
-
-Please refer to the [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
+Please refer to the Flow Aggregator Helm Chart [README](../build/charts/flow-aggregator/README.md)
+and [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
 to learn about more ClickHouse configuration options in Flow Aggregator.
 
+##### ClickHouse Cluster
+
+From Theia v0.2.0, we support deploying ClickHouse cluster with data replication.
```
jianjuns, line 220: a ClickHouse cluster

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+purpose, as when a Pod accidentally restarts, the memory will be cleared, which
+will lead to a complete data loss. In that case, please refer to the
+[ClickHouse doc](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication/#recovery-after-complete-data-loss)
+to recover the data.
+
+ClickHouse cluster can be deployed with default Local PV or NFS PV by setting
+`clickhouse.storage.createPersistentVolume`. Please make sure the provided path
+contains a folder for each replica with a name in the format `vol<i>`, where `i`
+varies from 0 to the total number of replicas. To have more flexibility in the
+PV creation, we recommend using customized StorageClass by setting
```
jianjuns, line 253: Do not understand the sentence. Do you mean: "To have more flexibility in the PV creation, you can configure a customized `StorageClass` in `clickhouse.storage.persistentVolumeClaimSpec`?

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+The default affinity allows only one ClickHouse instance per Node. Please make
+sure there are at least 4 Nodes available in your Kubernetes cluster to deploy
+a 2-shard-2-replica cluster with this affinity. To change the affinity, please
+set `clickhouse.cluster.podDistribution` per your requirement.
+
+We recommend using Persistent Volumes if you are going to use ClickHouse cluster
+in production. ClickHouse cluster with in memory deployment is mainly for test
+purpose, as when a Pod accidentally restarts, the memory will be cleared, which
+will lead to a complete data loss. In that case, please refer to the
+[ClickHouse doc](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication/#recovery-after-complete-data-loss)
```
jianjuns, line 246: doc -> document

docs/network-flow-visibility.md
yanjunz97: Thanks, updated.

docs/network-flow-visibility.md
yanjunz97: Thanks, updated.

docs/network-flow-visibility.md
yanjunz97: Yes. Updated. Thanks for the rephrasing!

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
-stringData:
-  password: clickhouse_operator_password
-  username: clickhouse_operator
-type: Opaque
-```
-
-Please refer to the [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
+Please refer to the Flow Aggregator Helm Chart [README](../build/charts/flow-aggregator/README.md)
+and [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
 to learn about more ClickHouse configuration options in Flow Aggregator.
```
ziyouw, line 217: To learn more about ClickHouse configuration optons?

docs/network-flow-visibility.md
```diff
@@ -202,31 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
 
+##### ClickHouse Cluster
+
+From Theia v0.2.0, we support deploying a ClickHouse cluster with data replication.
+A ClickHouse cluster consists of shards. Shards refer to the servers that contain
+different parts of the data, which provide a horizontal cluster scaling strategy.
+Each shard consists of one or more replica hosts to ensure reliability.
```
ziyouw, line 223: I think cluster mainly provides two things: shards for load balancing and replicas for HA. Maybe we should highlight the differences because users may only prefer one of them or both.

build/charts/theia/values.yaml
```diff
@@ -46,32 +47,68 @@ clickhouse:
...
     # volumeName: "<my-pv>".
     # To dynamically provision a PersistentVolume, please set
     # storageClassName: "<my-storage-class>".
-    # Memory storage is used if both createPersistentVolume.type and persistentVolumeClaimSpec are empty.
+    # Memory storage is used if both createPersistentVolume.type and
+    # persistentVolumeClaimSpec are empty.
     persistentVolumeClaimSpec: {}
       # storageClassName: ""
       # volumeName: ""
+  cluster:
```
ziyouw, line 81: For the cluster, how the monitoring script works? The problem is different instances may have differernt storage volumes.

docs/network-flow-visibility.md
yanjunz97: I think both make sense, but I do not have preference. Updated, thanks.

build/charts/theia/values.yaml
yanjunz97: The monitor will be deployed only on the last replica in each shard. It will check the local table on the shard and delete records when threshold is exceeded.

ziyouw: Does it mean each repilca in each shard should have the same size of storage volume?

yanjunz97: Yes, as each replica use the same `volumeClaimTemplates`, by default they should have the same storage size.

build/charts/theia/provisioning/datasources/create_table.sh
```diff
@@ -182,8 +207,13 @@ clickhouse client -n -h 127.0.0.1 <<-EOSQL
         sourcePodNamespace,
         destinationPodNamespace;
 
+    {{- if .Values.clickhouse.cluster.enable }}
```
salv-orlando, line 210: Looking at the way the DB is created we cannot avoid destroying/recreating the database when switching to clustered mode. (in theory we could have some complex logic for renaming old tables, creating new tables, and moving data around).

I think a strategy here could be to document that switching is not supported. Even if one toggled cluster.enable from False to True (or vice versa), we'd need to delete old tables to make sure the DB is properly created, which will work only for in-memory DB.
The other part of the strategy would to default to clustered mode, and possibly also deprecate non-clustered mode. At the end of the day, for a small devtest env it is enough to run a single CH replica, and a single ZK instance - is that correct?

build/charts/theia/README.md
```diff
@@ -18,6 +18,12 @@ Kubernetes: `>= 1.16.0-0`
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| clickhouse.cluster.enable | bool | `false` | Determine whether to deploy ClickHouse as a cluster. |
+| clickhouse.cluster.installZookeeper | object | `{"size":"5Gi"}` | Specification to install a default ZooKeeper cluster. This is ignored if zookeeperHosts are provided. The default ZooKeeper cluster is deployed with 3 replicas. |
+| clickhouse.cluster.podDistribution | list | `[{"number":1,"topologyKey":"kubernetes.io/hostname","type":"MaxNumberPerNode"}]` | Affinity for the ClickHouse Pods. By default, it allows only one ClickHouse instance per Node. Please refer to <https://github.com/Altinity/clickhouse-operator/blob/master/docs/chi-examples/99-clickhouseinstallation-max.yaml> for other distributions. |
+| clickhouse.cluster.replicas | int | `2` | Number of ClickHouse replicas in each shard. |
+| clickhouse.cluster.shards | int | `2` | Number of ClickHouse shards in the cluster. |
```
salv-orlando, line 25: I think the default config creates 2 shards with 2 replicas each.
Can we default to a single shard with 2 replicas, as that would be enough for many small deployments, and use less resources in our CI?

build/charts/theia/provisioning/datasources/create_table.sh
```diff
@@ -257,12 +291,36 @@ clickhouse client -n -h 127.0.0.1 <<-EOSQL
...
         yamls String
+    {{- if .Values.clickhouse.cluster.enable }}
+    ) engine=ReplicatedMergeTree('/clickhouse/tables/{shard}/{database}/{table}', '{replica}')
+    {{- else }}
     ) engine=MergeTree
+    {{- end }}
     ORDER BY (timeCreated);
-    
+
+    {{- if .Values.clickhouse.cluster.enable }}
```
salv-orlando, line 310: An idea about maintainability: can we add comment lines to the script explaining what we are doing in the various steps?

build/charts/theia/templates/clickhouse/zookeeper/statefulset.yaml
```diff
@@ -0,0 +1,182 @@
...
+                matchExpressions:
+                  - key: "app"
+                    operator: In
+                    values:
+                      - zookeeper
+              topologyKey: "kubernetes.io/hostname"
+      containers:
+        - name: kubernetes-zookeeper
+          imagePullPolicy: IfNotPresent
+          image: "docker.io/zookeeper:3.8.0"
```
salv-orlando, line 40: Shall we have an image hosted on our own dockerhub for the usual reasons?

build/charts/theia/templates/clickhouse/nfs-persistentvolume.yaml
```diff
@@ -1,16 +1,25 @@
...
+{{- end }}
+{{- range $i, $_ := until $pvNum }}
+{{- if not (eq $i 0) }}
+---
+{{- end }}
 apiVersion: v1
 kind: PersistentVolume
 metadata:
-  name: clickhouse-pv
+  name: clickhouse-pv-{{$i}}
```
salv-orlando, line 13: If I'm not mistaken, we are creating every DB on the same PV. What's the reasoning for this?

As a user I'm using replication for performance and fault tolerance. If all the replicas are on the same PV I hardly get any fault tolerance (for data loss), and also performance might not be that much different from single replica as all the I/O is still occurring on the same device.

Did you consider an option with a distinct PV per replica?

build/charts/theia/values.yaml
salv-orlando: What happens if the pod hosting the instance with the clickhouse monitor is destroyed for any reason? I assume it's reloaded again using the template with the monitor, is that correct?

Idea for a future enhancement: add leader election to the clickhouse monitor so that only an instance in a given shard will be active at any time.

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+the ClickHouse credentials, please update the `clickHouse.connectionSecret` in the
+Flow Aggregator Helm Chart values.
+
+Please refer to the Flow Aggregator Helm Chart [README](../build/charts/flow-aggregator/README.md)
+and [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
+to learn more about ClickHouse configuration options in Flow Aggregator.
+
+##### ClickHouse Cluster
+
+From Theia v0.2.0, we support deploying a ClickHouse cluster with data replication.
```
salv-orlando, line 220: suggestion: Starting with v0.2, Theia supports Clickhouse database cluster with data replication

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,56 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+replicas meta information. By default, a ZooKeeper cluster using `emptyDir` with
+3 replicas will be deployed for ClickHouse cluster. The 3 replicas are expected
+to be deployed on 3 different Nodes. To use a customized ZooKeeper cluster,
+please refer to the
+[ZooKeeper setup instructions for ClickHouse](https://github.com/Altinity/clickhouse-operator/blob/master/docs/zookeeper_setup.md)
+and set `clickhouse.cluster.zookeeperHosts` to your ZooKeeper hosts.
+
+The default affinity allows only one ClickHouse instance per Node. Please make
+sure there are at least 4 Nodes available in your Kubernetes cluster to deploy
+a 2-shard-2-replica cluster with this affinity. To change the affinity, please
```
salv-orlando, line 240: In the doc you did not mention before that the default values are 2 shard/2 replicas. It might be worth doing that as user might wonder why we're talking about this.

Also, due to this antiaffinity policy, can we have a single "local" PV for all instances?

build/charts/theia/values.yaml
yanjunz97: Yes, K8s will restart the pod with the right template containing the monitor. Thanks for the suggestion of leader election! I will investigate this later.

build/charts/theia/provisioning/datasources/create_table.sh
yanjunz97: It sounds good to add a devtest env first. Deploying single CH replica is already supported. I can add an option to deploy single ZK instance and doc this. But I'm not sure whether we should deprecate non-clustered mode, as it requires less resource by not deploying ZK, which might be useful for light-weighted env.

build/charts/theia/templates/clickhouse/nfs-persistentvolume.yaml
yanjunz97: To be clear, we are creating every replica on different PV, but all the PVs are on the same node for Local case, and the same NFS server for the NFS case. As users can use `persistentVolumeClaimSpec` for customized PVs, I plan to provide a simple case for the Local and NFS solution.

I'm thinking about if I introduce multiple nodes or NFS servers, then there might be cases that users do not have enough nodes or NFS servers, and they may want to have different PV numbers on each node or NFS servers, which makes it much more complex.

But using a distinct Node or NFS server per replica is not that complex. Your consideration about the fault tolerance and performance makes sense. It might be a better default case for our Local/NFS PV solution. Let me switch to this.

build/charts/theia/README.md
```diff
@@ -30,14 +36,14 @@ Kubernetes: `>= 1.16.0-0`
 | clickhouse.service.tcpPort | int | `9000` | TCP port number for the ClickHouse service. |
 | clickhouse.service.type | string | `"ClusterIP"` | The type of Service exposing ClickHouse. It can be one of ClusterIP, NodePort or LoadBalancer. |
 | clickhouse.storage.createPersistentVolume.local.affinity | object | `{}` | Affinity for the Local Persistent Volume. By default it requires to label the Node used to store the ClickHouse data with "antrea.io/clickhouse-data-node=". |
-| clickhouse.storage.createPersistentVolume.local.path | string | `"/data"` | The local path. Required when type is "Local". |
+| clickhouse.storage.createPersistentVolume.local.path | string | `"/data"` | The local path. Required when type is "Local". To use this Local PV for cluster, please create dir <path>/vol<i> on the Node, where i varies from 0 to shards * replicas. |
```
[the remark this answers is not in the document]
yanjunz97: Thanks for the suggestion. This sentence is removed as the PV creation strategy is changed. Updated in other comments.

build/charts/theia/README.md
```diff
@@ -30,14 +36,14 @@ Kubernetes: `>= 1.16.0-0`
...
 | clickhouse.storage.createPersistentVolume.nfs.host | string | `""` | The NFS server hostname or IP address. Required when type is "NFS". |
-| clickhouse.storage.createPersistentVolume.nfs.path | string | `""` | The path exported on the NFS server. Required when type is "NFS". |
+| clickhouse.storage.createPersistentVolume.nfs.path | string | `""` | The path exported on the NFS server. Required when type is "NFS". To use this NFS PV for cluster, please create dir <path>/vol<i> on the NFS server, where i varies from 0 to shards * replicas. |
```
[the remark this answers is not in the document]
yanjunz97: same as above, thanks!

build/charts/theia/provisioning/datasources/create_table.sh
yanjunz97: Updated to allow any number of ZK replica, and also the docs. Thanks!

build/charts/theia/README.md
yanjunz97: Got it. Update.

build/charts/theia/provisioning/datasources/create_table.sh
yanjunz97: Add some comments, thanks!

build/charts/theia/templates/clickhouse/zookeeper/statefulset.yaml
yanjunz97: Thanks @salv-orlando for pointing out this. I added the uploading workflow. Could you help to update the theia dockerhub token so that it can have the permission to upload the `theia-zookeeper` image? Thanks a lot!

build/charts/theia/templates/clickhouse/nfs-persistentvolume.yaml
yanjunz97: Now each replica will use distinct Node or NFS server.

docs/network-flow-visibility.md

docs/network-flow-visibility.md
yanjunz97: Updated, thanks! The PV problem is discussed and updated as https://github.com/antrea-io/theia/pull/55#discussion_r922450358

build/charts/theia/README.md
```diff
@@ -18,6 +18,14 @@ Kubernetes: `>= 1.16.0-0`
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| clickhouse.cluster.enable | bool | `false` | Determine whether to deploy ClickHouse as a cluster. |
```
salv-orlando, line 21: While we discuss on whether we consider non-clustered supported, do you think we can enable clustering by default?

build/charts/theia/README.md
```diff
@@ -30,14 +38,15 @@ Kubernetes: `>= 1.16.0-0`
...
+| clickhouse.storage.createPersistentVolume.local.nodes | list | `["kind-worker","kind-worker2"]` | A list of Node hostnames. Required when ClickHouse clustering is enabled and type is "Local". Please make sure to provide (shards * replicas) Nodes. Each Node should meet affinity and have the path created on it. |
 | clickhouse.storage.createPersistentVolume.local.path | string | `"/data"` | The local path. Required when type is "Local". |
-| clickhouse.storage.createPersistentVolume.nfs.host | string | `""` | The NFS server hostname or IP address. Required when type is "NFS". |
-| clickhouse.storage.createPersistentVolume.nfs.path | string | `""` | The path exported on the NFS server. Required when type is "NFS". |
+| clickhouse.storage.createPersistentVolume.nfs.hosts | list | `[]` | A list of NFS server hostname or IP address. Required when type is "NFS". When ClickHouse clustering is enabled, please provide (shards * replicas) NFS servers. Otherwise, only one NFS server is required. |
```
salv-orlando, line 43: I think we're not providing NFS servers but NFS shares, is that correct? They could be for instance all on the same server, mapped to locations assigned to different physical drives.

salv-orlando (commented):
Don't have much to add in terms of comments, as the change LGTM.
We need to finalize on our strategy regarding support for non-clustered Clickhouse and then we can finalize this PR.

build/charts/theia/README.md
yanjunz97: An NFS share could be defined as `nfs.hosts:nfs.path`. Do you think this is better to let user input a list of `nfs.hosts:nfs.path` instead of accepting the two values separately?

yanjunz97: Updated as above, thanks

build/charts/theia/README.md
ziyouw: +1

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,60 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+Starting with v0.2, Theia supports Clickhouse database cluster with data replication.
+A ClickHouse cluster consists of one or more shards. Shards refer to the servers
+that contain different parts of the data. You can deploy multiple shards to scale
+the cluster horizontally. Each shard consists of one or more replica hosts. You
+can deploy multiple replicas in a shard to ensure reliability.
+
+To deploy ClickHouse as a cluster, please set `clickhouse.cluster.enable` to true
+and set `clickhouse.cluster.shards` and `clickhouse.cluster.replicas` per your
+requirement. Please notice that a standlone ClickHouse server cannot switch to
+ClickHouse cluster without deleting. If you plan to scale the ClickHouse cluster
```
ziyouw, line 230: Should we mention that it also will lose data if user wants to switch a cluster to a standalone one?

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,60 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+##### ClickHouse Cluster
+
+Starting with v0.2, Theia supports Clickhouse database cluster with data replication.
+A ClickHouse cluster consists of one or more shards. Shards refer to the servers
+that contain different parts of the data. You can deploy multiple shards to scale
+the cluster horizontally. Each shard consists of one or more replica hosts. You
+can deploy multiple replicas in a shard to ensure reliability.
+
+To deploy ClickHouse as a cluster, please set `clickhouse.cluster.enable` to true
+and set `clickhouse.cluster.shards` and `clickhouse.cluster.replicas` per your
```
ziyouw, line 228: We just mainly explained what was shard, but replica is a new term now. Not sure if it is easy for users to understand the relationships between shard and replica.

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,60 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
-  namespace: flow-aggregator
-stringData:
-  password: clickhouse_operator_password
-  username: clickhouse_operator
-type: Opaque
-```
-
-Please refer to the [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
-to learn about more ClickHouse configuration options in Flow Aggregator.
+The ClickHouse TCP service is used by the Flow Aggregator. If you have changed
```
jianjuns, line 206: service -> Service

yanjunz97: Fixed, thanks!

docs/network-flow-visibility.md
yanjunz97: Added some explanations for replicas. Could you check if it is clearer now? Thank!

docs/network-flow-visibility.md
yanjunz97: It makes sense. Added, thanks!

build/charts/theia/README.md
yanjunz97: As we plan to deprecate the non-clustered, I think it might be good to remove `cluster.enable` here and always deploy ClickHouse cluster. Updated as this way.

yanjunz97:
@salv-orlando @ziyouw I have deprecated the non-cluster mode in this PR and verify the e2e tests. Would you like to take another look at this PR?

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,69 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+to learn more about ClickHouse configuration options in Flow Aggregator.
+
+##### ClickHouse Cluster
+
+In v0.1, Theia deploys ClickHouse in a non-cluster mode, which cannot switch to
+ClickHouse cluster directly. You need to uninstall ClickHouse, delete related
+data if you use PersistenVolume and redeploy it when upgrading or downgrading
+between v0.1 and v0.2.
+
+Starting with v0.2, Theia supports ClickHouse database cluster with data replication,
```
salv-orlando, line 226: nit: consider simplifying with replacing "ClickHouse database cluster with data replication" with "Clickhouse clustering" or "Clickhouse clustered deployment"

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,69 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+and deprecates the non-cluster mode. A ClickHouse cluster consists of one or more
```
salv-orlando, line 227: non-clustered mode is not deprecated, but removed.

Maybe we can simply state "Starting with v0.2, Theia supports ClickHouse clustering in lieu of the non-clustered mode supported in v0.1"

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,69 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+
+By default, Theia deploys a 1-shard-1-replica ClickHouse cluster and a 1-replica
+ZooKeeper. To scale the ClickHouse cluster, please set `clickhouse.cluster.shards`
+and `clickhouse.cluster.replicas` per your requirement.
+
+ClickHouse uses [Apache ZooKeeper](https://zookeeper.apache.org/) for storing
+replicas meta information. When scaling the ClickHouse cluster, please set
+`clickhouse.cluster.installZookeeper.replicas` to at lease 3 to ensure fault
+tolerance. Each replica is expected to be deployed on a different Node. To
+use a customized ZooKeeper cluster, please refer to the
```
salv-orlando, line 242: more than "customized ZK cluster", shall we consider "to leverage a user-provided ZK cluster..."?

salv-orlando (commented):
Awesome work, I just have some comments about documentation

docs/network-flow-visibility.md
yanjunz97: Updated as below, thanks!

docs/network-flow-visibility.md
yanjunz97: Sounds good. Updated, thanks!

docs/network-flow-visibility.md
yanjunz97: It's more accurate. Update. Thanks!

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,69 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+replicas meta information. When scaling the ClickHouse cluster, please set
+`clickhouse.cluster.installZookeeper.replicas` to at least 3 to ensure fault
+tolerance. Each replica is expected to be deployed on a different Node. To
+leverage a user-provided ZooKeeper cluster, please refer to the
+[ZooKeeper setup instructions for ClickHouse](https://github.com/Altinity/clickhouse-operator/blob/master/docs/zookeeper_setup.md)
+and set `clickhouse.cluster.zookeeperHosts` to your ZooKeeper hosts.
+
+When stopping the ClickHouse cluster, please ensure it is stopped before ZooKeeper,
+as ClickHouse relies on ZooKeeper to stop correctly. You can run the following
+command to stop the ClickHouse cluster.
```
yuntanghsu, line 248: What's the purpose of this paragraph? Is it for the users who want to switch their clickhouse back to non-cluster mode?

yanjunz97: No. Just for any user who wants to stop ClickHouse. I notice that if users stop the ClickHouse cluster with our manifests directly, it cannot stop correctly.

yuntanghsu: Thanks for explanation. Do you think we should combine this paragraph with L157 or move it before L157?

yanjunz97: I think it makes sense to mention this in the Deployment Section. But as this paragraph introduces ZooKeeper, I'm a little bit worry about adding these to L157 without the context will be confused. I try to update the Deployment Section. Please take a look if it is understandable.

For your question about the `helm uninstall`, yes, it is true.

yuntanghsu: It looks good now. Thanks.

salv-orlando (approved):

build/charts/theia/README.md
```diff
@@ -30,14 +37,14 @@ Kubernetes: `>= 1.16.0-0`
...
 | clickhouse.storage.createPersistentVolume.type | string | `""` | Type of PersistentVolume. Can be set to "Local" or "NFS". Please set this value to use a PersistentVolume created by Theia. |
-| clickhouse.storage.persistentVolumeClaimSpec | object | `{}` | Specification for PersistentVolumeClaim. This is ignored if createPersistentVolume.type is non-empty.  To use a custom PersistentVolume, please set  storageClassName: "" volumeName: "<my-pv>". To dynamically provision a PersistentVolume, please set storageClassName: "<my-storage-class>". Memory storage is used if both createPersistentVolume.type and persistentVolumeClaimSpec are empty. |
+| clickhouse.storage.persistentVolumeClaimSpec | object | `{}` | Specification for PersistentVolumeClaim. This is ignored if createPersistentVolume.type is non-empty. To use a custom PersistentVolume, please set  storageClassName: "" volumeName: "<my-pv>". To dynamically provision a PersistentVolume, please set storageClassName: "<my-storage-class>". Memory storage is used if both createPersistentVolume.type and persistentVolumeClaimSpec are empty. |
```
heanlan, line 44:
```suggestion
| clickhouse.storage.persistentVolumeClaimSpec | object | `{}` | Specification for PersistentVolumeClaim. This is ignored if createPersistentVolume.type is non-empty. To use a custom PersistentVolume, please set storageClassName: "" volumeName: "<my-pv>". To dynamically provision a PersistentVolume, please set storageClassName: "<my-storage-class>". Memory storage is used if both createPersistentVolume.type and persistentVolumeClaimSpec are empty. |
```

docs/network-flow-visibility.md
```diff
@@ -170,6 +171,15 @@ kubectl delete -f flow-visibility.yml
 kubectl delete -f https://raw.githubusercontent.com/antrea-io/theia/main/build/charts/theia/crds/clickhouse-operator-install-bundle.yaml -n kube-system
 ```
 
+From Theia 0.2, Theia supports ClickHouse clustering in lieu of the non-clustered
```
heanlan, line 174: Would it better if we put this paragraph before the instructions of stopping Grafana Flow Collector? As it is required to be done before those. No strong opinion here, would like to hear your thoughts.

heanlan (approved):

build/charts/theia/README.md
yanjunz97: Thanks Anlan for reviewing! Fixed.

docs/network-flow-visibility.md
yanjunz97: I agree. Slightly rephrase the paragraph. Would you like to take another look if it is better now?

heanlan: LGTM

heanlan (approved):
LGTM

yuntanghsu:
LGTM

yuntanghsu (approved):
LGTM

ziyouw (approved):

PR #63, opened 2022-06-23
Remove antctl checks from E2E test

heanlan:
This commit removes the antctl related checks from E2E test, as
the antctl source code are all in the Antrea repo. We also rename
flowaggregator_test to flowvisibility_test as it is more appropriate.


yanjunz97 (approved):
LGTM, only a nit. Maybe we can also remove this commented code?

https://github.com/antrea-io/theia/blob/32b5d7b19aea63224c07e3a5b9fc8b9d2fed5156/test/e2e/flowaggregator_test.go#L706

heanlan:
@yanjunz97 updated. Please help check.

test/e2e/flowvisibility_test.go
```diff
@@ -34,82 +34,76 @@ import (
...
+reverseOctetDeltaCountFromSourceNode: 7083284
+reverseOctetDeltaCountFromDestinationNode: 7083284
+reverseOctetTotalCountFromSourceNode: 24500996
+reverseOctetTotalCountFromDestinationNode: 24500996
+reversePacketDeltaCountFromSourceNode: 136211
+reversePacketDeltaCountFromDestinationNode: 136211
+reversePacketTotalCountFromSourceNode: 471111
+reversePacketTotalCountFromDestinationNode: 471111
+flowEndSecondsFromSourceNode: 1637706973
+flowEndSecondsFromDestinationNode: 1637706973
```
yanjunz97, line 97: How do we construct this sample? It still has some differences with the current data schema. For example, it seems that we have renamed `destinationClusterIPv4` to `destinationClusterIP`. We also removed all the fields  from `octetDeltaCountFromSourceNode` to `reversePacketTotalCountFromDestinationNode`.

heanlan: Originally it is sample output of IPFIX collector, which I think it's not appropriate to keep it anymore. Given there's no such a convenient way to generate the same output from clickhouse DB, do you think we'd better remove it?

yanjunz97: Not sure whether keeping it can provide convenience when verifying the records. But after taking a look I think a sample record in ClickHouse can be generated by running a query like
```
SELECT * FROM default.flows LIMIT 1 FORMAT Vertical
```

heanlan: thanks, updated

PR #69, opened 2022-07-01
Add Theia Helm chart to release

yanjunz97:

### DIFF
--- a/build/charts/Makefile
+++ b/build/charts/Makefile
@@ -1,6 +1,10 @@
 USERID  := $(shell id -u)
 GRPID   := $(shell id -g)
 
+VERSION := $(shell head -n 1 ../../VERSION | cut -c 2-)
+
 .PHONY: helm-docs
 helm-docs:
 	docker run --rm --volume "$(CURDIR):/helm-docs" --user=$(USERID):$(GRPID) jnorwood/helm-docs:v1.7.0
+	sed -i.bak "s/0.0.0/$(VERSION)/g" theia/README.md # replace version placeholder
+	sed -i.bak "s/-dev-informational/--dev-informational/g" theia/README.md # fix img.shields.io badge URLs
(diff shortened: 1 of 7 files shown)


build/charts/Makefile
```diff
@@ -1,6 +1,10 @@
 USERID  := $(shell id -u)
 GRPID   := $(shell id -g)
 
+VERSION := $(shell head -n 1 ../../VERSION | cut -c 2-)
+
 .PHONY: helm-docs
 helm-docs:
 	docker run --rm --volume "$(CURDIR):/helm-docs" --user=$(USERID):$(GRPID) jnorwood/helm-docs:v1.7.0
+	sed -i.bak "s/0.0.0/$(VERSION)/g" theia/README.md # replace version placeholder
```
dreamtalen, line 9: Looks like you also need to change the version in `theia/README.md` to 0.0.0?

yanjunz97: I'm not sure whether you refer to the inconsistency between the `Chart.yaml` and `README.md`. As `README.md` is not expected to be modified when we package the chart but version number in `Chart.yaml` is taken care of by the package command. Thus the version number in `README.md` is replaced when it is generated to be consistent with the version number of the repo instead of the version number of the `Chart.yaml`.

dreamtalen: I see, thanks for the explanation!

hack/generate-helm-release.sh
```diff
@@ -0,0 +1,93 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+function echoerr {
+    >&2 echo "$@"
+}
+
+_usage="Usage: $0 [--mode (dev|release)] --out <DIR>
```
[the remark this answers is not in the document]
yanjunz97: Making it the same as Antrea looks good to me. Updated, thanks!

PR #70, opened 2022-07-01
Remove the Flow Aggregator Helm Chart

yanjunz97:
Flow Aggregator Helm Chart is added to Antrea repo with PR https://github.com/antrea-io/antrea/pull/3952. This PR removes it from Theia repo.


docs/getting-started.md
```diff
@@ -41,24 +41,37 @@ Antrea deployment manifest:
 
 ## Theia Installation
 
+In Theia v0.1.0, we introduce the Flow Aggregator Helm Chart to Theia
+From Theia v0.2.0 and Antrea v1.8.0, the Flow Aggregator Helm Chart is moved from Theia to Antrea.
```
dreamtalen, line 45: Line wrapping.

docs/getting-started.md
```diff
@@ -41,24 +41,37 @@ Antrea deployment manifest:
...
 
-```bash
-git clone https://github.com/antrea-io/theia.git
-helm install flow-aggregator theia/build/charts/flow-aggregator --set clickHouse.enable=true -n flow-aggregator --create-namespace
-helm install theia theia/build/charts/theia -n flow-visibility --create-namespace
-```
+    From Theia v0.2.0 and Antrea v1.8.0, the Flow Aggregator Helm Chart is moved from
+    Theia repository to Antrea repository; and the Helm Charts are added to release
+    assets. Please download `theia-chart.tgz` from [Theia release assets](https://github.com/antrea-io/theia/releases)
+    and `flow-aggregator-chart.tgz` from [Antrea release assets](https://github.com/antrea-io/antrea/releases).
```
dreamtalen, line 59: Probably we also need to provide steps for unzipping these .tgz before installing

docs/getting-started.md
```diff
@@ -41,24 +41,37 @@ Antrea deployment manifest:
...
+    The Helm charts are located under the folder `build/charts`.
 
-```bash
-git clone https://github.com/antrea-io/theia.git
-helm install flow-aggregator theia/build/charts/flow-aggregator --set clickHouse.enable=true -n flow-aggregator --create-namespace
-helm install theia theia/build/charts/theia -n flow-visibility --create-namespace
-```
+    From Theia v0.2.0 and Antrea v1.8.0, the Flow Aggregator Helm Chart is moved from
+    Theia repository to Antrea repository; and the Helm Charts are added to release
+    assets. Please download `theia-chart.tgz` from [Theia release assets](https://github.com/antrea-io/theia/releases)
```
dreamtalen, line 58: I wonder if letting users download the helm charts is inconvenient.
When using other helm charts, they usually provide a `helm repo add` command first and then a `helm install` command.
May I ask what steps we need to have similar process?

docs/getting-started.md
yanjunz97: I forgot to remove these sentences after I rephrase them here:
https://github.com/antrea-io/theia/blob/3fa731ca4527c97fbfa0286d47590c89252e1faf/docs/getting-started.md?plain=1#L50-L58
Removed them. Please tell me if you think it is unclear.

docs/getting-started.md
yanjunz97: Make sense. Added, thanks!

docs/getting-started.md
yanjunz97: The way to turn our GitHub project into a self-hosted Helm chart repo I found is to use [helm/chart-releaser-action](https://github.com/helm/chart-releaser-action). But it is a little bit different in the release process. Instead of generating a chart release when we create the release from a branch, it automatically checks each push and releases the chart when the chart version number is changed. So currently I use the way Antrea repo uses to create the release, which simply packages the charts and uploads it.

antoninbas: Starting with Antrea v1.8.0, users will be able to do `helm repo add charts.antrea.io`. This is in progress. See https://github.com/antrea-io/website/pull/25.

dreamtalen: Got it, thanks Yanjun and Antonin.

yanjunz97: Have not noticed that. Thanks Antonin for the information!

yanjunz97: Hi @antoninbas I took a look at the PR quickly, it seems only the antrea helm chart is added. Could we also have the flow-aggregator chart from Antrea repo and theia chart from Theia repo added to the website?

antoninbas: Yes, but we are testing the mechanism end to end first. We can add the 2 remaining charts afterwards.

yanjunz97: Got it. Thanks!

salv-orlando (approved):
LGTM!

PR #81, opened 2022-07-23
Add ClickHouse data schema management

yanjunz97:
When ClickHouse data schema is updated in a new release, users using Persistent Volume
to deploy their ClickHouse will encounter the problem that the old data schema does not
match the new one, and have to lose all the old data.

To deal with this problem, this PR introduces version control to ClickHouse data schema, which
detects the data schema version when initializing ClickHouse.

An upgrading from version N-1 e2e test is also added in this PR.


yuntanghsu:
A possibly dumb question: I think our clickhouse is "in memory" by default in 0.1.0? If so, how do we keep the data when upgrade from/downgrade to 0.1.0?

yanjunz97:
> A possibly dumb question: I think our clickhouse is "in memory" by default in 0.1.0? If so, how do we keep the data when upgrade from/downgrade to 0.1.0?

Upgrading/downgrading is only designed for Persistent Volume. Although in 0.1 the default deployment is in memory, we have added the PV support. The data schema management is design for that cases.

build/charts/theia/provisioning/datasources/migrate.sh
```diff
@@ -0,0 +1,123 @@
...
+        dataVersion="0.1.0"
+    fi
+}
+
+function setDataVersion {
+    tables=$(clickhouse client -h 127.0.0.1 -q "SHOW TABLES")
+    if [[ $tables == *"migrate_version"* ]]; then
+        clickhouse client -h 127.0.0.1 -q "ALTER TABLE migrate_version DELETE WHERE version!=''"
+    else
+        clickhouse client -h 127.0.0.1 -q "CREATE TABLE migrate_version (version String) engine=MergeTree ORDER BY version"
```
salv-orlando, line 31: Does this need to be updated for a clustered database schema since a different engine needs to be used?

hack/generate-manifest.sh
```diff
@@ -140,6 +147,9 @@ fi
 if [ "$GRAFANA" == false ]; then
     HELM_VALUES+=("grafana.enable=false")
 fi
+if [[ $LOCALPATH != "" ]]; then
+    HELM_VALUES+=("clickhouse.storage.createPersistentVolume.type=Local" "clickhouse.storage.createPersistentVolume.local.path=$LOCALPATH")
+fi
```
salv-orlando, line 152: Is the change above related to this PR? At first glance I don't see how it's related to DB schema versioning

build/charts/theia/provisioning/datasources/migrate.sh
```diff
@@ -0,0 +1,123 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+function checkDataVersion {
+    tables=$(clickhouse client -h 127.0.0.1 -q "SHOW TABLES")
+    if [[ $tables == *"migrate_version"* ]]; then
+        dataVersion=$(clickhouse client -h 127.0.0.1 -q "SELECT version FROM migrate_version")
+    elif [[ $tables == *"flows"* ]]; then
```
salv-orlando, line 21: Do we need to explicitly check for *"flows"*?
(this is not important, I'm just trying to correctly understand the logic)

salv-orlando (commented):
Code looks mostly ok.

I guess that if we go for "upgrade to clustered model and only support that model" than the migration will not be a placeholder anymore but will need to include the code for updating the schema. We should probably assess how feasible that is.

I think it's fine to run schema management code in clickhouse init code, we don't need to use theia manager for this.

In the long run I think managing migrations with a shell script might become cumbersome.
While we can use this for 0.1 <-> 0.2 for the future I would consider different tooling.

golang: https://github.com/golang-migrate/migrate

or

python: https://github.com/sqlalchemy/alembic

(they both seem to support Clickhouse)

build/charts/theia/provisioning/datasources/migrate.sh
yanjunz97: I think it might be safe to keep as it is, as the table `migrate_version` is not expected to be changed after initialization.  During the initialization, each pod will run this schema management code to update the version independently.

hack/generate-manifest.sh
yanjunz97: To add e2e test for this schema management, we need to deploy ClickHouse with PersistentVolume. This change allows us to generate the YAML with PersistentVolume enabled during the e2e test setup.

build/charts/theia/provisioning/datasources/migrate.sh
yanjunz97: I think we need to check this to confirm whether there are data schema created before. When we find there is no `migrate_version` table, there might be two conditions:

- We are upgrading from v0.1. In this case, we can find the table `flows` in database. And we need to run the upgrading SQLs.
- We are deploying ClickHouse with an empty PersistentVolume. There is no data schema created before. In this case, we cannot find the table `flows`. And we can stop the migration and create the data schema.

build/charts/theia/provisioning/datasources/migrate.sh
```diff
@@ -0,0 +1,123 @@
...
+    elif [[ "$dataVersionIndex" -gt "$theiaVersionIndex" ]]; then
+        for i in $(seq $((dataVersionIndex-1)) -1 $theiaVersionIndex);
+        do
+            if test -f "$migratorDir/downgrade/${DOWNGRADING_MIGRATORS[$i]}"; then
+                echo "=== Apply file ${DOWNGRADING_MIGRATORS[$i]} ==="
+                clickhouse client -h 127.0.0.1 --queries-file $migratorDir/downgrade/${DOWNGRADING_MIGRATORS[$i]}
+            fi
+        done
+    else
+        echo "=== Data schema version is the same as ClickHouse version. Migration finished. ==="
```
wsquan171, line 120: same as Theia version?

build/charts/theia/provisioning/datasources/migrate.sh
```diff
@@ -0,0 +1,123 @@
...
+
+function migrate {
+    THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+    migratorDir="/var/lib/clickhouse/migrators"
+    mkdir -p $migratorDir
+
+    checkDataVersion
+    if [[ -z "$dataVersion" ]]; then
+        cp -r $THIS_DIR/migrators/* $migratorDir
+        echo "=== No existed data schema. Migration finished. ==="
```
wsquan171, line 88:
```suggestion
        echo "=== No existing data schema. Migration skipped. ==="
```

build/charts/theia/provisioning/datasources/init.sh
```diff
@@ -0,0 +1,26 @@
...
+set -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+source $THIS_DIR/migrate.sh
+source $THIS_DIR/create_table.sh
+
+migrate
+createTable
+setDataVersion
```
wsquan171, line 26: `setDataVersion` seems to be called already at the end of `migrate`. why calling again?

wsquan171 (commented):
Thanks Yanjun. mainly checked the migration part.

build/charts/theia/provisioning/datasources/migrate.sh
```diff
@@ -0,0 +1,123 @@
...
+        versionPair=(${fileName//_/ })
+        addVersionsToList ${versionPair[0]//-/.}
+        addVersionsToList ${versionPair[1]//-/.}
+    done
+    addVersionsToList $dataVersion
+    addVersionsToList $theiaVersion
+    # Sort the versions to generate the migration path
+    versionList=(${versionListStr//,/ })
+    IFS=$'\n' sortedversionList=($(sort -V <<<"${versionList[*]}"))
+    unset IFS
```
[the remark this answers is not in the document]
yanjunz97: Thanks Shawn for pointing out this! Fixed.

build/charts/theia/provisioning/datasources/migrate.sh
yanjunz97: Updated, thanks!

build/charts/theia/provisioning/datasources/migrate.sh
yanjunz97: Updated, thanks!

build/charts/theia/provisioning/datasources/init.sh
yanjunz97: Migration will be skipped if there is no data schema created before. In this case, I still want to set data version after running `createTable` function.

yanjunz97:
Hi @salv-orlando @ziyouw I've added the upgrading SQL from non-cluster to cluster mode. Would you like to take a look at it?

hack/generate-manifest.sh
```diff
@@ -35,6 +35,8 @@ Kustomize, and print it to stdout.
                                             Ei, Pi, Ti, Gi, Mi, Ki. (default is 8Gi)
         --ch-monitor-threshold <threshold>  Deploy the ClickHouse monitor with a specific threshold. Can
                                             vary from 0 to 1. (default is 0.5)
+        --local <path>                      Create the PersistentVolume for ClickHouse with a provided
```
salv-orlando, line 38: nit: ... for Clickhouse DB ...

hack/generate-manifest.sh
salv-orlando: understood!

build/charts/theia/provisioning/datasources/init.sh
salv-orlando: Then you can just set the version only at the end of the init process, so that it's executed regardless of whether the migration is skipped or not?

build/charts/theia/provisioning/datasources/migrate.sh
salv-orlando: Understood, so we could be checking for 'flows' as well as for _any_ table in the DB.
Makes sense.

salv-orlando (commented):
Good job! Some minor inline comments.

build/charts/theia/provisioning/datasources/init.sh
yanjunz97: I am thinking about a case that the migration is finished but the Pod crashes during `createTable`. In this way I think we may want to have the version updated, as we do not want to the migration part to run again after the recover. But this might be a rare case and might only happen in development when we have something wrong in the `createTable`. Please tell me if you think it is safe to ignore this case.

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,70 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+Please refer to the Flow Aggregator Helm Chart
+[README](https://github.com/antrea-io/antrea/blob/main/build/charts/flow-aggregator/README.md)
+and [Flow Aggregator Configuration](https://github.com/antrea-io/antrea/blob/main/docs/network-flow-visibility.md#configuration-1)
+to learn more about ClickHouse configuration options in Flow Aggregator.
+
+##### ClickHouse Cluster
+
+In v0.1, Theia deploys ClickHouse in a non-cluster mode. Starting with v0.2,
+Theia supports ClickHouse database cluster with data replication, and deprecates
+the non-cluster mode. You can upgrade from v0.1 to v0.2 without data losing if
```
dreamtalen, line 223: `data losing` -> `data loss` or `losing data`

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,70 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+you use PersistentVolume for ClickHouse. To downgrade from v0.2 to v0.1, you
+need to uninstall ClickHouse, delete related data if you use PersistenVolume
+and redeploy it.
+
+A ClickHouse cluster consists of one or more shards. Shards refer to the servers
+that contain different parts of the data. You can deploy multiple shards to scale
+the cluster horizontally. Each shard consists of one or more replica hosts.
+Replicas indicates storing the same data on multiple Pods to improve data
```
dreamtalen, line 231: `indicates` -> `indicate`

docs/network-flow-visibility.md
```diff
@@ -202,30 +203,70 @@ helm install theia build/charts/theia -n flow-visibility --create-namespace \
...
+availability and accessibility. You can deploy multiple replicas in a shard to
+ensure reliability.
+
+By default, Theia deploys a 1-shard-1-replica ClickHouse cluster and a 1-replica
+ZooKeeper. To scale the ClickHouse cluster, please set `clickhouse.cluster.shards`
+and `clickhouse.cluster.replicas` per your requirement.
+
+ClickHouse uses [Apache ZooKeeper](https://zookeeper.apache.org/) for storing
+replicas meta information. When scaling the ClickHouse cluster, please set
+`clickhouse.cluster.installZookeeper.replicas` to at lease 3 to ensure fault
```
dreamtalen, line 241: `at lease 3` -> `at least 3`

dreamtalen (commented):
Some nits in doc.

docs/network-flow-visibility.md

docs/network-flow-visibility.md
yanjunz97: This is introduced in the ClickHouse cluster PR. Updated there and rebased. Thanks!

docs/network-flow-visibility.md
yanjunz97: Also updated in the ClickHouse cluster PR and rebased. Thanks!

salv-orlando (approved):
LGTM!

dreamtalen (approved):
LGTM

yuntanghsu (approved):

ziyouw (approved):
LGTM.

PR #86, opened 2022-08-03
Update clickhouse server repo in GitHub workflow

wsquan171:
yandex/clickhouse-server is stale for 6 months now and the new repo is clickhouse/clickhouse-server:
https://hub.docker.com/r/clickhouse/clickhouse-server/tags

Given that current version we use (21.11) is no longer supported in term of security updates:
https://github.com/ClickHouse/ClickHouse/blob/master/SECURITY.md
this change updates the workflow with the new repo to allow updating to later versions of the server in case we want to. The new repo still contains the current tag.


dreamtalen (approved):
LGTM, thanks Shawn pointing out this.

PR #92, opened 2022-08-16
Add retry when sending record to clickhouse in theia_clickhouse_test

yuntanghsu:
For some reason, ClickHouse on jenkins-e2e-test will receive termination signal, which results in the disconnection with zookeeper. It takes time (~30s) for clickhouse to fix the error `error: Couldn't start replication (table will be in readonly mode)` by itself.

Adding retry to commit the query later.

### DIFF
--- a/test/e2e/theia_clickhouse_test.go
+++ b/test/e2e/theia_clickhouse_test.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"antrea.io/theia/pkg/theia/commands"
 )
@@ -96,18 +97,14 @@ const (
 	numFieldsInTableInfo = 7
 	dateBaseName         = "default"
 	defaultPath          = "/var/lib/clickhouse/"
+	flowsRelatedTableNum = 4
 )
 
 var tableColumnNumberMap = map[string]string{
-	".inner.flows_node_view":         "16",
 	".inner.flows_node_view_local":   "16",
-	".inner.flows_pod_view":          "20",
 	".inner.flows_pod_view_local":    "20",
-	".inner.flows_policy_view":       "27",
 	".inner.flows_policy_view_local": "27",
-	"flows":                          "49",
 	"flows_local":                    "49",
-	"recommendations":                "4",
 	"recommendations_local":          "4",
 }
 
@@ -186,13 +183,13 @@ func testTheiaGetClickHouseDiskInfo(t *testing.T, data *TestData) {
 // 1              default        flows                    50000          13.09 MiB      49
 // 1              default        recommendations          10             2.34 KiB       4
 func testTheiaGetClickHouseTableInfo(t *testing.T, data *TestData, connect *sql.DB) {
-	// send 10000 records to clickhouse
-	commitNum := 10
+	// send 1000 records to clickhouse
+	commitNum := 1
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		sendTraffic(t, commitNum, connect)
+		sendTraffic(t, connect, commitNum)
 	}()
 	wg.Wait()
 	// retrieve metrics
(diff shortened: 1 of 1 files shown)


yanjunz97:
May I learn about why do you choose to increase the ZK replica number instead of decrease the commit interval? It is a little bit strange to me that we need to deploy 3 ZK replicas to support only 1 ClickHouse replica. And I also notice the flow visibility e2e test is always stable which may indicate single-replica ZK can handle the flows when insertion rate is under some value? BTW, I'm not sure if increasing the ZK memory size would help. The default value is 5GB, but I think we can increase this default value if it helps.

yuntanghsu:
> May I learn about why do you choose to increase the ZK replica number instead of increase the commit interval? It is a little bit strange to me that we need to deploy 3 ZK replicas to support only 1 ClickHouse replica. And I also notice the flow visibility e2e test is always stable which may indicate single-replica ZK can handle the flows when insertion rate is under some value? BTW, I'm not sure if increasing the ZK memory size would help. The default value is 5GB, but I think we can increase this default value if it helps.

The problem is that, for some reason, jenkins-e2e test is not passed every time. And I notice that, it might because of the limited resource on jenkins' node. So the zookeeper will keep disconnecting when we run the test and making the table in readonly mode. 

`Error:      	Received unexpected error:
        	            	code: 242, message: Table is in readonly mode (zookeeper path: /clickhouse/tables/0/default/.inner.flows_node_view_local)`

test/e2e/theia_clickhouse_test.go
```diff
@@ -90,12 +91,13 @@ const (
...
 	recordPerCommit      = 1000
 	insertInterval       = 1
-	threshold            = 25
+	threshold            = 35
 	MaxInt32             = 1<<31 - 1
 	numFieldsInDiskInfo  = 9
 	numFieldsInTableInfo = 7
 	dateBaseName         = "default"
 	defaultPath          = "/var/lib/clickhouse/"
+	flowsRelatedTableNum = 4
```
dreamtalen, line 100: May I ask which four tables are related?

test/e2e/theia_clickhouse_test.go
```diff
@@ -270,8 +272,8 @@ func testTheiaGetClickHouseInsertRate(t *testing.T, data *TestData, connect *sql
 		assert.Equal(4, len(tableInfoArray), "tableInfoArray: %s", tableInfoArray)
 		actualInsertRate, error := strconv.Atoi(tableInfoArray[1])
 		assert.NoError(error)
-		tableNum := len(tableColumnNumberMap)
-		percent := (actualInsertRate/tableNum - recordPerCommit/insertInterval) * 100 / (recordPerCommit / insertInterval)
+		percent := (actualInsertRate/flowsRelatedTableNum - recordPerCommit/insertInterval) * 100 / (recordPerCommit / insertInterval)
+		percent = int(math.Abs(float64(percent)))
 		assert.LessOrEqualf(percent, threshold, "stdout: %s, expectedInsertRate: %s", stdout, recordPerCommit/insertInterval)
```
dreamtalen, line 277: I think these 3 lines of code could be simplified using `assert.InDeltaf`.
Reference: we are comparing the throughput difference should be less than 15% in https://github.com/antrea-io/theia/blob/main/test/e2e/flowvisibility_test.go#L760

yanjunz97:
> > May I learn about why do you choose to increase the ZK replica number instead of increase the commit interval? It is a little bit strange to me that we need to deploy 3 ZK replicas to support only 1 ClickHouse replica. And I also notice the flow visibility e2e test is always stable which may indicate single-replica ZK can handle the flows when insertion rate is under some value? BTW, I'm not sure if increasing the ZK memory size would help. The default value is 5GB, but I think we can increase this default value if it helps.
> 
> The problem is that, for some reason, jenkins-e2e test is not passed every time. And I notice that, it might because of the limited resource on jenkins' node. So the zookeeper will keep disconnecting when we run the test and making the table in readonly mode.
> 
> `Error: Received unexpected error: code: 242, message: Table is in readonly mode (zookeeper path: /clickhouse/tables/0/default/.inner.flows_node_view_local)`

Do you mean even if the insertion rate is low, we still get this error in ClickHouse CLI test? I'm just curious about why we do not see the similar failure in TestPolicyRecommendation or TestFlowVisibility, as I think these two tests should consume more resources than TestTheiaClickHouseStatusCommand. Both of these tests deploy extra components like Grafana and Spark, but the ZooKeeper has not been disconnected in these two tests.

test/e2e/theia_clickhouse_test.go
yuntanghsu: inner.flows_node_view_local, inner.flows_pod_view_local, inner.flows_policy_view_loca, flows_local

yuntanghsu:
> Do you mean even if the insertion rate is low, we still get this error in ClickHouse CLI test? I'm just curious about why we do not see the similar failure in TestPolicyRecommendation or TestFlowVisibility, as I think these two tests should consume more resources than TestTheiaClickHouseStatusCommand. Both of these tests deploy extra components like Grafana and Spark, but the ZooKeeper has not been disconnected in these two tests.

I tried to use {recordPerCommit=4000, insertInterval=8} and still saw the issue sometimes. And the reason I used such high recordPerCommit is because the insertionRate in clickhouse is for all tables (not only for the 4 flows tables). It will have around 160 rows/s even there is not traffic sent.

test/e2e/theia_clickhouse_test.go
yuntanghsu: Thanks for the suggestion.

test/e2e/theia_clickhouse_test.go
```diff
@@ -350,23 +345,35 @@ func addFakeRecord(t *testing.T, stmt *sql.Stmt) {
...
-	tx, err := connect.Begin()
-	require.NoError(t, err)
-	stmt, _ := tx.Prepare(insertQuery)
-	defer stmt.Close()
-	for j := 0; j < recordPerCommit; j++ {
-		addFakeRecord(t, stmt)
-	}
-	err = tx.Commit()
-	assert.NoError(t, err)
+	err := wait.PollImmediate(5*time.Second, 90*time.Second, func() (bool, error) {
```
dreamtalen, line 348: Better naming some constants for `5*time.Second` and `90*time.Second`

dreamtalen (approved):
overall LGTM

yanjunz97 (approved):
LGTM

PR #95, opened 2022-08-24
Adopt Golang version ClickHouse data schema management

yanjunz97:
We introduce the ClickHouse data schema management code implemented
in shell script in v0.2.0. In terms of the future maintenance, this PR
re-implements this function in Golang with the library [golang-migrate](https://github.com/golang-migrate/migrate).


plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,326 @@
...
+		return fmt.Errorf("error when getting Theia version: %v", err)
+	}
+	dataVersionNumber, err := getDataVersionNumber(*clickhouseMigrate)
+	if err != nil {
+		return fmt.Errorf("error when getting the data version: %v", err)
+	}
+	if theiaVersionNumber == dataVersionNumber {
+		klog.InfoS("Data schema version is the same as Theia version. Migration finished.")
+	} else if dataVersionNumber == -1 {
+		klog.InfoS("No existed data schema. Migration skipped.")
```
yuntanghsu, line 96:
```suggestion
		klog.InfoS("No existing data schema. Migration skipped.")
```

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,326 @@
...
+		// fileName example: 000001_0-1-0.down.sql
+		fileNameArr := strings.Split(strings.Split(file.Name(), ".")[0], "_")
+		versionNumber, err := strconv.ParseInt(fileNameArr[0], 10, 64)
+		if err != nil {
+			return fmt.Errorf("error when parsing the version number: %v", err)
+		}
+		version := strings.Replace(fileNameArr[1], "-", ".", -1)
+		// File <golang-migrate-version>_<theia-version>.up.sql is expected to be
+		// applied when upgrading from <theia-version>.
+		// golang-migrate apply this file when upgrading from <golang-migrate-version> - 1.
```
yuntanghsu, line 146:
```suggestion
		// golang-migrate applies this file when upgrading from <golang-migrate-version> - 1.
```

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,326 @@
...
+			continue
+		}
+		return (xi < yi), nil
+	}
+	return false, nil
+}
+
+// go-migrate is introduced in Theia v0.3. It cannot distinguish versions before v0.3 automatically.
+// If table flows_local is created, the data version is v0.2.
+// If only table flows_local is not existed while table flows is created, the data version is v0.1.
```
yuntanghsu, line 242:
```suggestion
// If only table flows_local doesn't exist while table flows is created, the data version is v0.1.
```
or 
```suggestion
// If only table flows_local is not present while table flows is created, the data version is v0.1.
```
?

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,326 @@
...
+// If non of these two tables are found, it means not data schema has been created before.
```
yuntanghsu, line 243:
```suggestion
// If non of these two tables are found, it means no data schema has been created before.
```

ci/kind/test-e2e-kind.sh
```diff
@@ -170,7 +170,7 @@ function run_test {
   rm -rf $TMP_DIR
   sleep 1
 
-  go test -v -timeout=30m antrea.io/theia/test/e2e -provider=kind --logs-export-dir=$ANTREA_LOG_DIR --skip=$skiplist
+  # go test -v -timeout=30m antrea.io/theia/test/e2e -provider=kind --logs-export-dir=$ANTREA_LOG_DIR --skip=$skiplist
```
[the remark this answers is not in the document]
yanjunz97: Fixed, thanks!

yuntanghsu (approved):
LGTM

yanjunz97:
As one of the reason we bump ClickHouse to v22.6 is that it supports both `amd64` and `arm64`, I add a commit to keep this feature when we build our ClickHouse server image.

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,338 @@
...
+	err = clickhouseMigrate.Force(theiaVersionNumber)
+	if err != nil {
+		return fmt.Errorf("error when setting version: %v", err)
+	}
+	return nil
+}
+
+func copyMigrators() error {
+	bashCmd := fmt.Sprintf("mkdir -p %s", migratorPersistentPath)
+	cmd := execCommand("bash", "-c", bashCmd)
```
salv-orlando, line 115: Could we use os.Mkdir (or os.MkdirAll) ?

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,338 @@
...
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error when creating folder: %s, error: %v", bashCmd, err)
+	}
+	bashCmd = fmt.Sprintf("cp -r %s/* %s", migratorTmpPath, migratorPersistentPath)
```
salv-orlando, line 119: While I do not love shelling out, I agree that using os.read & os.write would be a bit cumbersome.

My only advice would be - if possible - to not open a shell, and distinguish command "cp" from arguments.
Eg.:

execCommand("cp", fmt.Sprintf(" -r %s/* %s", migratorTmpPath, migratorPersistentPath))

As migratorTmpPath/migratorPersistentPath are constants there's no need to sanitize them.

These measures might not be needed here but it's always good to implement measures to prevent shell injection (some more info here: https://semgrep.dev/docs/cheat-sheets/go-command-injection/)

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,338 @@
...
+		}
+		if less && (versionNumber < value+1) {
+			versionNumber = value + 1
+		}
+	}
+	return versionNumber, nil
+}
+
+// Return true if version a is earlier than version b.
+func versionLessThan(a, b string) (bool, error) {
```
salv-orlando, line 232: Does this work also with a version number like 0.3.1?

(if I understand correctly this is the Theia version number, not the DB schema version number)

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,338 @@
...
+		if file.IsDir() {
+			continue
+		}
+		// fileName example: 000001_0-1-0.down.sql
+		fileNameArr := strings.Split(strings.Split(file.Name(), ".")[0], "_")
+		versionNumber, err := strconv.ParseInt(fileNameArr[0], 10, 64)
+		if err != nil {
+			return fmt.Errorf("error when parsing the version number: %v", err)
+		}
+		version := strings.Replace(fileNameArr[1], "-", ".", -1)
```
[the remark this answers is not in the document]
yanjunz97: I think we do not assume the order. Each file is named as `<number>_<theia_version>.up.sql` or `<number>_<theia_version>.down.sql`. And we are construct a map with Theia version as the key and the number as the value.

plugins/clickhouse-schema-management/main.go
yanjunz97: Yes, this function splits the Theia version 0.3.1 to (0, 3, 1) and compares the number in the tuple from left to right. In this logic, 0.3.1 is later than 0.3.0 but earlier than 0.4.0.

build/charts/theia/provisioning/datasources/migrators/000001_0-1-0.up.sql
```diff
@@ -0,0 +1,84 @@
...
+    reverseThroughputFromSourceNode UInt64,
+    reverseThroughputFromDestinationNode UInt64,
+    trusted UInt8 DEFAULT 0
+) engine=ReplicatedMergeTree('/clickhouse/tables/{shard}/{database}/{table}', '{replica}')
+ORDER BY (timeInserted, flowEndSeconds)
+TTL timeInserted + INTERVAL {{ .Values.clickhouse.ttl }}
+SETTINGS merge_with_ttl_timeout = {{ $ttlTimeout }};
+
+--Move data from old table and drop old tables
+INSERT INTO flows_local SELECT * FROM flows;
```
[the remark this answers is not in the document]
yanjunz97: In most cases, the answer is no, as `flows` will be a distributed table have references of `flows_local` on each replica. Most modules will be able to use `flows` as the interface to read/write data. This change is introduced when we introduce the non-cluster mode, and we have already update the code in other modules that are influenced by this change.

plugins/clickhouse-schema-management/main.go
```diff
@@ -0,0 +1,338 @@
...
+		// File <golang-migrate-version>_<theia-version>.up.sql is expected to be
+		// applied when upgrading from <theia-version>.
+		// golang-migrate applies this file when upgrading from <golang-migrate-version> - 1.
+		versionMap[version] = int(versionNumber - 1)
+	}
+	return nil
+}
+
+// Get Theia version based on the environment variable
+func getTheiaVerstionNumber() (int, error) {
```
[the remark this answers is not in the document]
yanjunz97: Fixed, thanks!

plugins/clickhouse-schema-management/main.go
yanjunz97: It makes sense for me. Updated, thanks!

plugins/clickhouse-schema-management/main.go
yanjunz97: Use `os.MkdirAll`, thanks!

build/charts/theia/templates/_helpers.tpl
```diff
@@ -66,13 +78,13 @@
       - key: {{ regexReplaceAll "(.*)/" $path "" }}
         path: {{ regexReplaceAll "(.*)/" $path "" }}
       {{- end }}
-      {{- range $path, $_ :=  $Files.Glob  "provisioning/datasources/migrators/upgrade/*" }}
+       {{- range $path, $_ :=  $Files.Glob  "provisioning/datasources/migrators/downgrade/*" }}
```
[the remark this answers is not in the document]
yanjunz97: It's an extra space. Removed, thanks!

plugins/clickhouse-schema-management/main.go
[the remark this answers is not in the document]
salv-orlando: cool, I double checked and there is no possibility for an entry in the map to be overridden with an incorrect value if the filenames are reported in a random order by readDir

plugins/clickhouse-schema-management/main.go
salv-orlando: My bad - I misread the code in my previous review!

salv-orlando (approved):
LGTM!

build/charts/theia/provisioning/datasources/migrators/000001_0-1-0.up.sql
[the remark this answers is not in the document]
salv-orlando: This is also not new code, it's the migration code from 0.1 to 0.2. It shows up in this diff because it has been moved into a different place.

PR #99, opened 2022-08-24
Update Go to v1.19

antoninbas:
To stay aligned with antrea-io/antrea

Most notables changes:
* we remove all uses of the io/ioutil package, which has been deprecated
* gofmt now reformats doc comments (e.g., lists inside comment blocks),
  which is why some comments have been reformatted automatically
* we now use the version of Go from the root go.mod file when setting up
  Go in Github workflows. This will reduce the number of changes we have
  to do in the future when updating Go.

Signed-off-by: Antonin Bas <abas@vmware.com>


plugins/clickhouse-monitor/main.go
```diff
@@ -60,6 +64,21 @@ var (
 	monitorExecInterval time.Duration
 )
 
+var notAValidIdentifierError = errors.New("not a valid identifier")
+
+func sanitizeIdentifier(identifier string) (string, error) {
```
dreamtalen, line 69: Clickhouse identifiers cannot be equal to keywords either, maybe worth adding a comment that we only check the regex here.
https://clickhouse.com/docs/en/sql-reference/syntax/#identifiers

test/e2e/policyrecommendation_test.go
```diff
@@ -255,10 +255,12 @@ func testPolicyRecommendationFailed(t *testing.T, data *TestData) {
 }
 
 // Example output:
-// apiVersion: crd.antrea.io/v1alpha1
-// kind: NetworkPolicy
-// metadata:
-//   name: recommend-allow-anp-fj3hd
+//
+//	apiVersion: crd.antrea.io/v1alpha1
```
dreamtalen, line 259: I assume this change is just for better formatting, as there is no such indentation in the output.

dreamtalen (approved):
LGTM overall

test/e2e/policyrecommendation_test.go
antoninbas: it's mentioned in the PR description
https://tip.golang.org/doc/go1.19#go-doc

plugins/clickhouse-monitor/main.go
antoninbas: sure, I'll add a comment. it's unpractical to check for keywords here, and there is no risk of SQl injection with a single keyword (no space), it will be rejected by ClickHouse

antoninbas: done

heanlan (approved):

yuntanghsu (approved):

yanjunz97 (approved):

PR #100, opened 2022-08-25
Fix failures in the clair scan test

yanjunz97:
This PR fixes failures in the clair scan test with the following changes:

- clair-scan seems to have problem in analyzing images building from
scratch. Removed the scan for ClickHouse monitor image in this test.
- AWS keys are missing in Theia repo, but they are only used for email
notification. Removed the keys and email sending step.
- Removed scanning for the ClickHouse images with constant tag, as
we do not build and update them.


dreamtalen (approved):
LGTM, thanks Yanjun fixing this.

PR #102, opened 2022-08-29
Improvement for throughput related panels in Grafana

yuntanghsu:
Change the throughput calculation for the panels aggregating connections' throughput.

Some panels use sum(throughput) when grouping the connections. The throughput it shows is not quite correct. For some short connections, they will occupy a large of proportion when we do the calculation.

Change it to SUM(octetDeltaCount)*8000/$__interval_ms to have accurate result.

Affected panel:
1. Throughput of Pod as Source (pod-to-pod-dashboard)
2. Throughput of Pod as Destination (pod-to-pod-dashboard)
3. Throughput of Node-to-Node (node-to-node-dashboard)
4. Reverse Throughput of Node-to-Node (node-to-node-dashboard)
5. Throughput of Node as Source (node-to-node-dashboard)
6. Throughput of Node as Destination (node-to-node-dashboard)
7. Throughput of Pod as Source (pod-to-service-dashboard)
8. Throughput of Pod as Destination (pod-to-service-dashboard)

### DIFF
--- a/build/charts/theia/provisioning/dashboards/node_to_node_dashboard.json
+++ b/build/charts/theia/provisioning/dashboards/node_to_node_dashboard.json
@@ -22,7 +22,7 @@
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
   "id": 3,
-  "iteration": 1659133528716,
+  "iteration": 1661796451586,
   "links": [],
   "liveNow": false,
   "panels": [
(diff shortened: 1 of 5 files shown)


build/charts/theia/provisioning/dashboards/node_to_node_dashboard.json
```diff
@@ -426,7 +426,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT $__timeInterval(flowEndSecondsFromSourceNode) as time, sourceNodeName, SUM(throughputFromSourceNode)\nFROM flows_node_view\nWHERE sourceNodeName != '' \nAND destinationNodeName != ''\nAND sourcePodNamespace NOT IN ('kube-system', 'flow-visibility', 'flow-aggregator')\nAND destinationPodNamespace NOT IN ('kube-system', 'flow-visibility', 'flow-aggregator')\nAND $__timeFilter(time)\nGROUP BY time, sourceNodeName\nORDER BY time",
+          "rawSql": "SELECT $__timeInterval(flowEndSecondsFromSourceNode) as time, sourceNodeName, SUM(octetDeltaCount)*8000/$__interval_ms\nFROM flows_node_view\nWHERE sourceNodeName != '' \nAND destinationNodeName != ''\nAND sourcePodNamespace NOT IN ('kube-system', 'flow-visibility', 'flow-aggregator')\nAND destinationPodNamespace NOT IN ('kube-system', 'flow-visibility', 'flow-aggregator')\nAND $__timeFilter(time)\nGROUP BY time, sourceNodeName\nORDER BY time",
```
heanlan, line 429: Since we are using common field - `octetDeltaCount`, instead of node specific field - `throughputFromSourceNode `. To make the time field and stats field consistent, I would suggest to use `flowEndSeconds` here.

yuntanghsu: Thanks for the suggestion, I've updated it.

dreamtalen (approved):
LGTM

PR #87, opened 2022-08-05
Bump ClickHouse image from 21.11 to 22.6

wsquan171:
Signed-off-by: Shawn Wang <wshaoquan@vmware.com>

PR #103, opened 2022-08-29
Updated theia-clickhouse-server image for arm64 support

Dhruv-J:
This commit adds updated directions to network-flow-visibility.md for deploying
through standalone manifest. Arm64 architecture requires an updated image for
theia-clickhouse-server.

Signed-off-by: Dhruv Jain <dhruvj@vmware.com>

### DIFF
--- a/build/yamls/flow-visibility.yml
+++ b/build/yamls/flow-visibility.yml
@@ -5215,7 +5215,7 @@ spec:
         - env:
           - name: THEIA_VERSION
             value: 0.3.0
-          image: projects.registry.vmware.com/antrea/theia-clickhouse-server:21.11
+          image: projects.registry.vmware.com/antrea/theia-clickhouse-server:22.6
           imagePullPolicy: IfNotPresent
           name: clickhouse
           volumeMounts:


dreamtalen (commented):
Thanks Dhruv for the PR.
In another way, we could update `theia-clickhouse-server` to the v22.6 in Theia. @yanjunz97 Do you have any concerns about this?

yanjunz97:
> Thanks Dhruv for the PR. In another way, we could update `theia-clickhouse-server` to the v22.6 in Theia. @yanjunz97 Do you have any concerns about this?

If we can ensure that the current functionalities will not be broken, setting the default `theia-clickhouse-server` to v22.6 sounds to be a better solution for me.

Dhruv-J:
> > Thanks Dhruv for the PR. In another way, we could update `theia-clickhouse-server` to the v22.6 in Theia. @yanjunz97 Do you have any concerns about this?
> 
> If we can ensure that the current functionalities will not be broken, setting the default `theia-clickhouse-server` to v22.6 sounds to be a better solution for me.

Yeah, I can see if all the tests work, and make a different PR for that instead.

yanjunz97:
> > > Thanks Dhruv for the PR. In another way, we could update `theia-clickhouse-server` to the v22.6 in Theia. @yanjunz97 Do you have any concerns about this?
> > 
> > 
> > If we can ensure that the current functionalities will not be broken, setting the default `theia-clickhouse-server` to v22.6 sounds to be a better solution for me.
> 
> Yeah, I can see if all the tests work, and make a different PR for that instead.

Thanks Dhruv! You can just do the update in this PR.

wsquan171:
we had another PR trying to do the same with #87. previous e2e testing was successful.

Dhruv-J:
> we had another PR trying to do the same with #87. previous e2e testing was successful.

Sounds great, will close this one.

PR #104, opened 2022-09-01
Fix placeholder errors in ClickHouse monitor

yanjunz97:
`clickhouse-go/v1` does not support `$` as the placeholder and
leads to errors in e2e tests.

As we pull ClickHouse monitor image from Harbor to run e2e tests,
this bug is not detected when it is introduced. This PR also updates
the workflow to build the ClickHouse monitor image for kind/jenkins
e2e tests.

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -45,10 +45,29 @@ jobs:
         name: policy-recommendation
         path: policy-recommendation.tar
         retention-days: 1 # minimum value, in case artifact deletion by 'artifact-cleanup' job fails
+  
+  build-clickhouse-monitor-image:
+    name: Build ClickHouse monitor image to be used for Kind e2e tests
+    needs: check-changes
+    if: ${{ needs.check-changes.outputs.has_changes == 'yes' }}
+    runs-on: [ ubuntu-latest ]
+    steps:
+    - uses: actions/checkout@v3
+    - run: make clickhouse-monitor
+    - name: Save ClickHouse monitor image to tarball
+      run: docker save -o clickhouse-monitor.tar antrea/theia-clickhouse-monitor
+    - name: Upload ClickHouse monitor image for subsequent jobs
+      uses: actions/upload-artifact@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/kind.yml
```diff
@@ -127,12 +161,21 @@ jobs:
...
   artifact-cleanup:
     name: Delete uploaded images
-    needs: [build-policy-recommendation-image, test-e2e-encap]
-    if: ${{ always() && needs.build-policy-recommendation-image.result == 'success' }}
+    needs:
+    - build-policy-recommendation-image
+    - build-clickhouse-monitor-image
+    - test-e2e-encap
+    - test-upgrade-from-N-1
+    if: ${{ always() && (needs.build-policy-recommendation-image.result == 'success' || needs.build-clickhouse-monitor-image.result == 'success') }}
```
wsquan171, line 169: why using or between policy-recommendation and clickhouse-monitor build results?

yanjunz97: Because the cleanup job is expected to run if any of the images is successfully built.

yuntanghsu (commented):
LGTM

yuntanghsu (approved):

.github/workflows/kind.yml
dreamtalen, line 169: Looks like Quan has made an update on the Antrea side that:
`artifact-cleanup job now runs only when all test jobs succeed, to be
able to re-run failed jobs without re-running the image building job
and the succeeded jobs that depend on it. After all jobs succeed, the
cleanup job will still run, or the artifact will be deleted after
retention days (currently set to 1 day).`
I think we could follow this change in Theia.
Ref: https://github.com/antrea-io/antrea/pull/3985

yanjunz97: Make sense to me. Updated, thanks!

dreamtalen (approved):

wsquan171 (approved):
LGTM

PR #107, opened 2022-09-12
Unify using dollar sign for shell commands in CLI doc

dreamtalen:
Signed-off-by: Yongming Ding <dyongming@vmware.com>


yuntanghsu (approved):
LGTM

PR #109, opened 2022-09-14
Add networkpolicyrecommendations API definitions

wsquan171:
This change improves intelligence/networkpolicyrecommendations REST body definition and includes job status.

Signed-off-by: Shawn Wang <wshaoquan@vmware.com>


pkg/apis/intelligence/v1alpha1/types.go
```diff
@@ -31,19 +41,32 @@ type NetworkPolicyRecommendation struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type NetworkPolicyRecommendationStatus struct {
+	State                    JobState `json:"state,omitempty"`
+	SparkApplication         string   `json:"sparkApplication,omitempty"`
+	CompletedStages          int      `json:"completedStages,omitempty"`
+	TotalStages              int      `json:"totalStages,omitempty"`
+	RecommendedNetworkPolicy string   `json:"recommendedNetworkPolicy,omitempty"`
+	ErrorCode                string   `json:"errorCode,omitempty"`
+	ErrorMsg                 string   `json:"errorMsg,omitempty"`
```
yanjunz97, line 69: Just to confirm some points:

1. Does SparkApplication refer to the job ID in the current CLI?
2. Does RecommendedNetworkPolicy here only refer to YAMLs text in the recommendation?
3. Where does `ErrorCode` come from? I assume ErrorMsg contains the errors generated in NetworkPolicyRecommendation or API server, or propagated from SparkApplication. Is this something we will define only in the API server side? 
4. Can I assume that we can have a similar NetworkPolicyRecommendationStatus at CRD side but
    - It does not have the field `ErrorCode`
    - It has RecommendedNetworkPolicy as a struct for RecommendedNetworkPolicy CRD instead of string (I do not have a strong preference on this. String can also be a good choice as we do not have use cases for other fields)

pkg/apis/intelligence/v1alpha1/types.go
```diff
@@ -31,19 +41,32 @@ type NetworkPolicyRecommendation struct {
...
-	ExecutorInstances   int         `json:"executorInstances,omitempty"`
-	DriverCoreRequest   string      `json:"driverCoreRequest,omitempty"`
-	DriverMemory        string      `json:"driverMemory,omitempty"`
-	ExecutorCoreRequest string      `json:"executorCoreRequest,omitempty"`
-	ExecutorMemory      string      `json:"executorMemory,omitempty"`
+	Type                JobType                           `json:"jobType,omitempty"`
+	Limit               int                               `json:"limit,omitempty"`
+	PolicyType          string                            `json:"policyType,omitempty"`
+	StartTime           metav1.Time                       `json:"startTime,omitempty"`
+	EndTime             metav1.Time                       `json:"endTime,omitempty"`
```
yanjunz97, line 48: Aa Salvatore suggested in Slack, I plan to rename these two fields at CRD side to `IntervalStart` and `IntervalEnd` to reduce the confusion that these two look like the start and end time of NetworkPolicyRecommendation CR. I'm not sure whether we should also update here or if you have other suggestions on the naming.

yanjunz97 (commented):
Thanks Shawn for working on this!

pkg/apis/intelligence/v1alpha1/types.go
wsquan171: the new naming sounds good to me. will also update doc.

pkg/apis/intelligence/v1alpha1/types.go
wsquan171: 1. it could be id. I'm thinking that given the relationship between Job CR and Application CR we may want to convey that association in the status as well. 
2. for v0.3, yes. I also updated confluence page for details
3. error codes are to be convey theia manager application level errors back to client. i.e. spec validation failure, spark application creation failure, other unexpected errors out of SparkApplication context, etc. So that client (CLI) can programatically response based on error code rather than the error message
4. We can adapt based on need. I can't think of a example use case of ErrorCode at the moment. but for RecommendedNetworkPolicy in status I feel that reference should be better. Otherwise we'll have duplicate data in both job CR and that reco-policy CR

yuntanghsu: Thanks Shawn.

I'm also thinking we might need a similar or same struct of NetworkPolicyRecommendation in apiserver and crd.
Since the NPR CR will be updated in the controller, I think it might be helpful if we can update these status there?
Besides, we can directly copy and paste the data between apiserver and crd?

pkg/apis/intelligence/v1alpha1/types.go
```diff
@@ -16,11 +16,14 @@ package v1alpha1
...
 const (
-	NPRecommendationJobInitial    JobType = "Initial"
-	NPRecommendationJobSubsequent JobType = "Subsequent"
+	NPRecommendationJobInitial    string = "Initial"
+	NPRecommendationJobSubsequent string = "Subsequent"
+	NPRecommendationJobNew        string = "NEW"
+	NPRecommendationJobScheduled  string = "SCHEDULED"
+	NPRecommendationJobRunning    string = "RUNNING"
+	NPRecommendationJobCompleted  string = "COMPLETED"
+	NPRecommendationJobFailed     string = "FAILED"
```
yanjunz97, line 26: Maybe we can remove `Job` in these names? I think `Job` in `NPRecommendationJobInitial` refers to the job type. But in terms of state, it is confused to still have the word `Job` in the name?

yuntanghsu (approved):
LGTM

pkg/apis/intelligence/v1alpha1/types.go
salv-orlando: Or also StartInterval and EndInterval... really makes no difference to me, I prefer having field semantically similar close when reading them in alphabetical order, but other might prefer having "Start" and "End" at the beginning as it immediately provides detailed semantic info.

pkg/apis/intelligence/v1alpha1/types.go
```diff
@@ -31,19 +34,30 @@ type NetworkPolicyRecommendation struct {
...
+	ExecutorMemory      string                            `json:"executorMemory,omitempty"`
+	Status              NetworkPolicyRecommendationStatus `json:"status,omitempty"`
+}
+
+type NetworkPolicyRecommendationStatus struct {
+	State                    string `json:"state,omitempty"`
+	SparkApplication         string `json:"sparkApplication,omitempty"`
+	CompletedStages          int    `json:"completedStages,omitempty"`
+	TotalStages              int    `json:"totalStages,omitempty"`
+	RecommendedNetworkPolicy string `json:"recommendedNetworkPolicy,omitempty"`
```
salv-orlando, line 58: pedant nit: I suggest "RecommendeNetworkPolicies" (plural) or "RecommendationOutcome"

pkg/apis/intelligence/v1alpha1/types.go
```diff
@@ -31,19 +34,30 @@ type NetworkPolicyRecommendation struct {
...
+	ExecutorCoreRequest string                            `json:"executorCoreRequest,omitempty"`
+	ExecutorMemory      string                            `json:"executorMemory,omitempty"`
+	Status              NetworkPolicyRecommendationStatus `json:"status,omitempty"`
+}
+
+type NetworkPolicyRecommendationStatus struct {
+	State                    string `json:"state,omitempty"`
+	SparkApplication         string `json:"sparkApplication,omitempty"`
+	CompletedStages          int    `json:"completedStages,omitempty"`
+	TotalStages              int    `json:"totalStages,omitempty"`
```
salv-orlando, line 57: Would it be possible to add completion timestamps? (doesn't need to be done now, we can add them later)

salv-orlando (approved):

PR #111, opened 2022-09-16
Implement rest handler for apiserver

yuntanghsu:
Add NetworkPolicyRecommnedation rest handler
Add unit-test for rest.go


pkg/querier/querier.go
```diff
@@ -16,8 +16,12 @@ package querier
...
 	"antrea.io/theia/pkg/apis/crd/v1alpha1"
+	intelligence "antrea.io/theia/pkg/apis/intelligence/v1alpha1"
 )
 
 type NPRecommendationQuerier interface {
-	GetNetworkPolicyRecommendation(namespace, name string) (*v1alpha1.NetworkPolicyRecommendation, error)
+	GetNetworkPolicyRecommendation(name string) (*intelligence.NetworkPolicyRecommendation, error)
+	CreateNetworkPolicyRecommendation(*v1alpha1.NetworkPolicyRecommendation) (*v1alpha1.NetworkPolicyRecommendation, error)
+	DeleteNetworkPolicyRecommendation(name string) error
+	ListNetworkPolicyRecommendation() (*intelligence.NetworkPolicyRecommendationList, error)
```
yanjunz97, line 26: Thanks Yunyang for working on this and opening this PR early. I know you might still work on this PR, but I'm not sure if you plan to change this. Anyway, I think it might be helpful if we could reach consensus on this part before integration.

The definitions here look a little bit strange to me. IMO, the controller we implement should be a general controller, which does not have dependencies on intelligence apis or a specific namespace. Any work like reading results from database or deleting results should be handled by the handlers instead of the querier functions.

I would prefer the `NPRecommendationQuerier` interface defined as follows.
```golang
type NPRecommendationQuerier interface {
	GetNetworkPolicyRecommendation(namespace, name string) (*v1alpha1.NetworkPolicyRecommendation, error)
	ListNetworkPolicyRecommendation(namespace string) ([]*v1alpha1.NetworkPolicyRecommendation, error)
	DeleteNetworkPolicyRecommendation(namespace, name string) error
	CreateNetworkPolicyRecommendation(namespace string, networkPolicyRecommendation *v1alpha1.NetworkPolicyRecommendation) (*v1alpha1.NetworkPolicyRecommendation, error)
}
```

With a very simple implementation in controller as

```golang
func (c *NPRecommendationController) GetNetworkPolicyRecommendation(namespace, name string) (*crdv1alpha1.NetworkPolicyRecommendation, error) {
	return c.npRecommendationLister.NetworkPolicyRecommendations(namespace).Get(name)
}

func (c *NPRecommendationController) ListNetworkPolicyRecommendation(namespace string) ([]*crdv1alpha1.NetworkPolicyRecommendation, error) {
	return c.npRecommendationLister.NetworkPolicyRecommendations(namespace).List(labels.Everything())
}

func (c *NPRecommendationController) DeleteNetworkPolicyRecommendation(namespace, name string) error {
	return c.crdClient.CrdV1alpha1().NetworkPolicyRecommendations(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
}

func (c *NPRecommendationController) CreateNetworkPolicyRecommendation(namespace string, networkPolicyRecommendation *crdv1alpha1.NetworkPolicyRecommendation) (*crdv1alpha1.NetworkPolicyRecommendation, error) {
	return c.crdClient.CrdV1alpha1().NetworkPolicyRecommendations(namespace).Create(context.TODO(), networkPolicyRecommendation, metav1.CreateOptions{})
}
```
Or we could also include some generic options when defining the Interface as the followings. But I'm not sure if this is necessary as I have not come up with a use case of this based on our current requirements. 

```golang
type NPRecommendationQuerier interface {
	GetNetworkPolicyRecommendation(namespace, name string) (*v1alpha1.NetworkPolicyRecommendation, error)
	ListNetworkPolicyRecommendation(namespace string, selector labels.Selector) ([]*v1alpha1.NetworkPolicyRecommendation, error)
	DeleteNetworkPolicyRecommendation(namespace, name string, opts metav1.DeleteOptions) error
	CreateNetworkPolicyRecommendation(namespace string, networkPolicyRecommendation *v1alpha1.NetworkPolicyRecommendation, opts metav1.CreateOptions) (*v1alpha1.NetworkPolicyRecommendation, error)
}
```

Would also like to include @wsquan171 in this discussion to help us reach consensus on the `NPRecommendationQuerier` Interface definition. And I'm not quite sure if it is better to include these in PR #109

yuntanghsu: Thanks Yanjun for reviewing.
This PR is not finished yet. I was planning to open this and have some discussion with you and Shawn.
Your suggestions make sense to me. I also think update the status in the controller is not a good idea.

I did that because I was thinking our controller might be the place to update the Status field (SparkApplication status sync, Update RecommendNP). Since the NetworkPolicyRecommendation struct is different in intelligence api and crd, I think the controller might have the resource we need for NetworkPolicyRecommendationStatus.

I'm not sure if the handler will do duplicate job if it need to access the status data from SparkApplication and ClickHouse. Also, I'm curious that, since the Status field includes a recommendedNP field, do I still need to fetch this data from ClickHouse?

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
```diff
@@ -95,3 +89,74 @@ func (r *REST) NamespaceScoped() bool {
...
+	job.Spec.ExcludeLabels = npReco.ExcludeLabels
+	job.Spec.ToServices = npReco.ToServices
+	job.Spec.ExecutorInstances = npReco.ExecutorInstances
+	job.Spec.DriverCoreRequest = npReco.DriverCoreRequest
+	job.Spec.DriverMemory = npReco.DriverMemory
+	job.Spec.ExecutorCoreRequest = npReco.ExecutorCoreRequest
+	job.Spec.ExecutorMemory = npReco.ExecutorMemory
+	_, err := r.npRecommendationQuerier.CreateNetworkPolicyRecommendation(defaultNameSpace, job)
+	if err != nil {
+		return nil, errors.NewBadRequest(fmt.Sprintf("error when create NetworkPolicyRecommendation CR: %v", err))
```
yanjunz97, line 119:
```suggestion
		return nil, errors.NewBadRequest(fmt.Sprintf("error when creating NetworkPolicyRecommendation CR: %v", err))
```

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
```diff
@@ -95,3 +89,74 @@ func (r *REST) NamespaceScoped() bool {
...
+	}
+	return &metav1.Status{Status: metav1.StatusSuccess}, nil
+}
+
+func (r *REST) Delete(_ context.Context, name string, _ rest.ValidateObjectFunc, _ *metav1.DeleteOptions) (runtime.Object, bool, error) {
+	_, err := r.npRecommendationQuerier.GetNetworkPolicyRecommendation(defaultNameSpace, name)
+	if err != nil {
+		return nil, false, errors.NewBadRequest(fmt.Sprintf("networkPolicyRecommendation job doesn't exist, name: %s", name))
```
yanjunz97, line 127:
```suggestion
		return nil, false, errors.NewBadRequest(fmt.Sprintf("NetworkPolicyRecommendation job doesn't exist, name: %s", name))
```

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
```diff
@@ -95,3 +89,74 @@ func (r *REST) NamespaceScoped() bool {
...
+	intelli.DriverMemory = crd.Spec.DriverMemory
+	intelli.ExecutorCoreRequest = crd.Spec.ExecutorCoreRequest
+	intelli.ExecutorMemory = crd.Spec.ExecutorMemory
+	intelli.Status.State = crd.Status.State
+	intelli.Status.SparkApplication = crd.Status.SparkApplication
+	intelli.Status.CompletedStages = crd.Status.CompletedStages
+	intelli.Status.TotalStages = crd.Status.TotalStages
+	intelli.Status.RecommendedNetworkPolicy = crd.Status.RecommendedNP.Spec.Yamls
+	intelli.Status.ErrorCode = crd.Status.ErrorMsg
+	// todo: need to parse the error code
```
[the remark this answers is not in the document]
yuntanghsu: Fixed.
I haven't come up with a good definition and use case of ErrorCode.

pkg/apis/intelligence/v1alpha1/types.go
```diff
@@ -16,11 +16,14 @@ package v1alpha1
...
 const (
-	NPRecommendationJobInitial    JobType = "Initial"
-	NPRecommendationJobSubsequent JobType = "Subsequent"
+	NPRecommendationJobInitial    string = "Initial"
+	NPRecommendationJobSubsequent string = "Subsequent"
+	NPRecommendationJobNew        string = "NEW"
+	NPRecommendationJobScheduled  string = "SCHEDULED"
+	NPRecommendationJobRunning    string = "RUNNING"
+	NPRecommendationJobCompleted  string = "COMPLETED"
+	NPRecommendationJobFailed     string = "FAILED"
```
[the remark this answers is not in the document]
yuntanghsu: They shouldn't be here as they are already defined in crd's type. I've removed it. Thanks.

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
yuntanghsu: Fixed. Thanks.

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
yuntanghsu: Fixed. Thanks.

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
```diff
@@ -47,44 +53,32 @@ func (r *REST) New() runtime.Object {
 	return &intelligence.NetworkPolicyRecommendation{}
 }
 
-func (r *REST) getNetworkPolicyRecommendation(name string) *intelligence.NetworkPolicyRecommendation {
-	npReco, err := r.npRecommendationQuerier.GetNetworkPolicyRecommendation("flow-visibility", name)
+func (r *REST) Get(_ context.Context, name string, _ *metav1.GetOptions) (runtime.Object, error) {
```
wsquan171, line 56: nit: not sure if we should follow convention in Antrea and keep the unused arg name rather than using `_`, so that when later we add code using them, the arg names will be naturally kept same as the interface value.

wsquan171 (commented):
overall lgtm. one nit.

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
yuntanghsu: Thanks Shawn for reviewing. I've changed it to follow the style on Antrea side.

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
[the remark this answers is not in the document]
yanjunz97: I'm fine to remove ErrorCode if it does not have a use case. Would like to get @wsquan171 's thought on this.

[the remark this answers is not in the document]
wsquan171: Originally the error code was added to facilitate API clients to programmatically react on different types of errors based on error code instead of error msg. For instance, a front end may prompt user differently if the error is by nature invalid input or cluster availability issue. error codes are in general better way to traceback issues in workflows compared to error msg.

For current CLI, since there's little internal retry, and propagating error msg back to stdout is good enough, the use of error code may not be that obvious. For long term serviceability considerations, I still think it's a good-to-have. cc @salv-orlando for more feedback on this.

wsquan171 (approved):
LGTM

pkg/controller/networkpolicyrecommendation/controller_test.go
```diff
@@ -663,3 +694,82 @@ func TestGetPolicyRecommendationResult(t *testing.T) {
 		})
 	}
 }
+
+func initTestObjects(t *testing.T, stopCh chan struct{}) *NPRecommendationController {
```
yanjunz97, line 698: Maybe remove the code in this commit for controller unit test. It seems these unit tests overlap with the existed ones and do not increase the coverage.

yuntanghsu: Yes, they are duplicated. I've removed it. Thanks.

yanjunz97:
Overall LGTM, just think we need to have a decision on how to deal with the error code before merging.

yanjunz97 (approved):

PR #112, opened 2022-09-16
Introduce Snowflake backend for Theia

antoninbas:
Snowflake can be used as the data store for Theia, and to run applications such as policy recommendation (using Snowflake virtual warehouses). Users need to bring their own Snowflake and AWS accounts.

In this first commit, we add a new CLI tool (theia-sf), which can be used to create the necessary cloud resources in Snowflake and AWS. The CLI is built using Pulumi, an "infrastructure-as-code" tool, which is itself built using Terraform providers. No third-party tool needs to be installed directly by users, as theia-sf will install dependencies itself in a temporary directory.

When running "theia-sf onboard", which will provision infrastructure, users need to:
* configure AWS and Snowflake credentials
* provide an S3 bucket to store infrastructure state
* provide a KMS key to encrypt infrastructure state (notably secrets)

The advantage of using Pulumi to provision cloud resources is the potential ability to support upgrades easily: resources can be created, updated or deleted as needed whenever "theia-sf onboard" runs.

Pulumi is used to create a Snowflake database and schema. However, it is not used to create objects (tables, views, etc.) within that schema. For that we use a database migration tool. The rationale is that the Pulumi / Terraform provider for Snowflake doesn't support table schema updates very well, and our current solution is more flexible. Once again, this should enable us to support upgrades easily.

Signed-off-by: Antonin Bas <abas@vmware.com>


snowflake/database/migrations/000002_create_pods_view.up.sql
```diff
@@ -0,0 +1,35 @@
+CREATE VIEW IF NOT EXISTS pods (
```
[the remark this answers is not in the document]
antoninbas: - I think it's nicer to keep them separate (single statement per file)
- the view depends on the table
- multi-statement is not supported, so I would need to wrap them in a transaction (haven't tried to see if it works)

snowflake/cmd/offboard.go
```diff
@@ -0,0 +1,91 @@
...
+	rootCmd.AddCommand(offboardCmd)
+
+	offboardCmd.Flags().String("region", GetEnv("AWS_REGION", defaultRegion), "region where bucket should be created")
+	offboardCmd.Flags().String("stack-name", "default", "Name of the infrastructure stack: useful to deploy multiple instances in the same Snowflake account or with the same bucket (e.g., one for dev and one for prod)")
+	offboardCmd.Flags().String("bucket-name", "", "bucket to store infra state")
+	offboardCmd.MarkFlagRequired("bucket-name")
+	offboardCmd.Flags().String("bucket-prefix", "antrea-flows-infra", "prefix to use to store infra state")
+	offboardCmd.Flags().String("bucket-region", "", "region where infra bucket is defined; if omitted, we will try to get the region from AWS")
+	offboardCmd.Flags().String("key-id", GetEnv("THEIA_SF_KMS_KEY_ID", ""), "Kms Key ID")
+	offboardCmd.Flags().String("key-region", "", "Kms Key region")
```
dreamtalen, line 89:
```suggestion
	offboardCmd.Flags().String("key-region", "", "Kms key region")
```
to be consistent with elsewhere

snowflake/cmd/onboard.go
```diff
@@ -0,0 +1,128 @@
...
+	rootCmd.AddCommand(onboardCmd)
+
+	onboardCmd.Flags().String("region", GetEnv("AWS_REGION", defaultRegion), "region where AWS resources will be provisioned")
+	onboardCmd.Flags().String("stack-name", "default", "name of the infrastructure stack: useful to deploy multiple instances in the same Snowflake account or with the same bucket (e.g., one for dev and one for prod)")
+	onboardCmd.Flags().String("bucket-name", "", "bucket to store infra state")
+	onboardCmd.MarkFlagRequired("bucket-name")
+	onboardCmd.Flags().String("bucket-prefix", "antrea-flows-infra", "prefix to use to store infra state")
+	onboardCmd.Flags().String("bucket-region", "", "region where infra bucket is defined; if omitted, we will try to get the region from AWS")
+	onboardCmd.Flags().String("key-id", GetEnv("THEIA_SF_KMS_KEY_ID", ""), "Kms key ID")
+	onboardCmd.Flags().String("key-region", "", "Kms Key region")
```
dreamtalen, line 125:
```suggestion
	onboardCmd.Flags().String("key-region", "", "Kms key region")
```
to be consistent with elsewhere

snowflake/cmd/receiveSqsMessage.go
```diff
@@ -0,0 +1,121 @@
...
+	Short: "Receive a message from an AWS SQS queue",
+	Long: `This command can be used to receive a message from an SQS queue
+in your AWS account. Snowflake data ingestion errors are sent to an SQS queue,
+whose ARN is displayed when running the "onboard" command. While ingestion
+errors are not expected when using the Antrea Flow Aggregator to export flows,
+this command can prove useful if flow records don't get ingested into your
+Snowflake account as expected.
+
+To receive a message without deleting it (message will remain in the queue and
+become available to consummers again after a short time interval):
```
dreamtalen, line 42:
```suggestion
become available to consumers again after a short time interval):
```

snowflake/cmd/receiveSqsMessage.go
```diff
@@ -0,0 +1,121 @@
...
+		MaxNumberOfMessages: int32(1),
+		WaitTimeSeconds:     int32(0),
+	})
+	if err != nil {
+		return fmt.Errorf("error when receiving message from SQS queue: %v", err)
+	}
+	if len(output.Messages) == 0 {
+		return nil
+	}
+	message := output.Messages[0]
```
dreamtalen, line 99: Looks like we could only see the top message using this command. Shall we have an option that could display all messages?

dreamtalen (commented):
partial reviews

snowflake/README.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+### Configure AWS credentials
+
+Follow the steps in the [AWS
+documentation](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials)
+to specify credentials that can be used by the AWS Go SDK. Either configure the
+~/.aws/credentials file or set the required environment variables.
+
+You can also export the `AWS_REGION` environment variable, set to your preferred
```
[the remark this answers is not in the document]
antoninbas: this is fine I think, "set" is also the past participle of the "set" verb

snowflake/cmd/receiveSqsMessage.go
antoninbas: Honestly even though this is not difficult to do, I am not sure this is needed. We can see if this is requested in the future.

heanlan:
for dashboard needs, can we add a few fields to the `pods` view:
```sql
CREATE VIEW IF NOT EXISTS pods(
  flowStartSeconds,
  flowEndSeconds,
  packetDeltaCount,
  octetDeltaCount,
  reversePacketDeltaCount,
  reverseOctetDeltaCount,
  sourcePodName,
  sourcePodNamespace,
  source,
  destinationPodName,
  destinationPodNamespace,
  destination,
  throughput,
  reverseThroughput,
  clusterUUID,
  flowType,
  timeInserted
) as SELECT
  flowStartSeconds,
  flowEndSeconds,
  packetDeltaCount,
  octetDeltaCount,
  reversePacketDeltaCount,
  reverseOctetDeltaCount,
  sourcePodName,
  sourcePodNamespace,
  sourcePodNamespace || '/' || sourcePodName || ':' || SOURCETRANSPORTPORT,
  destinationPodName,
  destinationPodNamespace,
  destinationPodNamespace || '/' || destinationPodName|| ':' || DESTINATIONTRANSPORTPORT,
  throughput,
  reverseThroughput,
  flowtype,
  timeInserted
 FROM flows
 ```
 
 and add another `policies` view:
 ```sql
 CREATE VIEW IF NOT EXISTS policies 
 as SELECT
  flowEndSeconds,
  octetDeltaCount,
  reverseOctetDeltaCount,
  egressNetworkPolicyName,
  egressNetworkPolicyNamespace,
  egressNetworkPolicyRuleAction,
  ingressNetworkPolicyName,
  ingressNetworkPolicyNamespace,
  ingressNetworkPolicyRuleAction,
  sourcePodName,
  sourcePodNamespace,
  sourceTransportPort,
  destinationIP,
  destinationPodName,
  destinationPodNamespace,
  destinationTransportPort,
  destinationServicePortName,
  destinationServicePort,
  throughput,
  flowtype
 FROM flows
 ```
 
 thanks

**Comment by antoninbas** [comment]
@heanlan I need a clarification for one of the changes to the `pods` view. Is it necessary to add the source transport port to the `source` field? I am asking because usually information about the source port is not very important. Is this used anywhere in the dashboards? And more generally, does it make sense to concatenate the port number with namespace + name, or should we keep it as a separate field?

**Comment by heanlan** [comment]
> I need a clarification for one of the changes to the `pods` view. Is it necessary to add the source transport port to the `source` field? I am asking because usually information about the source port is not very important. Is this used anywhere in the dashboards? And more generally, does it make sense to concatenate the port number with namespace + name, or should we keep it as a separate field?

In ClickHouse dashboards, we use `sourcePodNamespace/sourcePodName:port -> destinationPodNamespace/destinationPodName:port` to identify a pod-to-pod connection. Because sometimes a sourcePod restarts, and its port changes. We think they should be considered as two different connections before&after the restart.

In ClickHouse, the [schema](https://github.com/antrea-io/theia/blob/c9ef257a0afbd4a6d77cc30d340a006039aa928c/build/charts/theia/provisioning/datasources/create_table.sh#L87) of view does not concatenate any fields. We do the concatenation in the queries. Because we use the same view in pod-to-pod, pod-to-service, pod-to-external dashboards. The fields need to be concatenated are be different according to the dashboards. Currently in snowflake, I didn't build pod-to-service, pod-to-external dashboards, and I saw you already have created concatenated fields `source` and `destination`, so I was thinking maybe it's ok to also concatenate port.

**Comment by antoninbas** [comment]
@heanlan if this is only useful for one dashboard and you can do it in the query, it may not make sense to concatenate the port in the view at the moment. I concatenate namespace + name in the view because it is useful for filtering IMO, independently of the dashboard.

**Comment by antoninbas** [comment]
@heanlan could you review the SQL migrations changes
- I kept the `clusterUUID` in the `pods` view (you had removed it in your comment)
- I added the `clusterUUID` to the `policies` view

Could you also confirm that you don't need `timeInserted` in the `policies` view?

File: snowflake/database/migrations/000002_create_pods_view.up.sql [hunk 7]
```diff
@@ -30,6 +31,7 @@ CREATE VIEW IF NOT EXISTS pods (
   destinationPodNamespace || '/' || destinationPodName,
   throughput,
   reverseThroughput,
+  flowtype,
```
heanlan, line 34: but we still want to add the `sourceTransportPort` and `destinationTransportPort` as two separate fields to the view?

antoninbas: ok I can do that

antoninbas: done

heanlan:
Sorry, I missed this one. 

> Could you also confirm that you don't need timeInserted in the policies view?

don't need timeInserted in policies

> Let me know if you think we can remove timeInserted from the pods view as well. Maybe we don't need it, I am not sure why I added it in the first place.

don't need timeInserted in pods

antoninbas:
@heanlan thanks for confirming, I removed `timeInserted` from the pods view

snowflake/database/migrations/000002_create_pods_view.up.sql
```diff
@@ -0,0 +1,39 @@
...
+  sourceTransportPort,
+  sourcePodNamespace || '/' || sourcePodName,
+  destinationPodName,
+  destinationPodNamespace,
+  destinationTransportPort,
+  destinationPodNamespace || '/' || destinationPodName,
+  throughput,
+  reverseThroughput,
+  flowtype,
+  clusterUUID,
```
heanlan, line 38:
```suggestion
  clusterUUID
```

snowflake/database/migrations/000003_create_policies_view.up.sql
```diff
@@ -0,0 +1,24 @@
...
+ sourceTransportPort,
+ destinationIP,
+ destinationPodName,
+ destinationPodNamespace,
+ destinationTransportPort,
+ destinationServicePortName,
+ destinationServicePort,
+ throughput,
+ flowtype,
+ clusterUUID,
```
heanlan, line 23:
```suggestion
 clusterUUID
```

heanlan (commented):
Sorry for adding comments incrementally. I believe it should be the last comment on the schema part.

antoninbas:
> Sorry for adding comments incrementally. I believe it should be the last comment on the schema part.

no worries, it's my bad for having these typos
I'm planning to run a new round of e2e tests after we go through review and before merging

dreamtalen (approved):
LGTM

heanlan (approved):
LGTM

PR #114, opened 2022-09-20
Add NetworkPolicyRecommendation Controller implementations

yanjunz97:
This PR implements the controller workflow for NetworkPolicyRecommendation
and adds RecommendedNetworkPolicy CRD to store the results.

The implementation of the controller is ready for review. Unit tests will be added later.

To run the NetworkPolicyRecommendation, please follow the steps below:

1. Deploy ClickHouse, Spark Operator, and Theia Manager.
2. Apply the following yaml. 

```yaml
apiVersion: crd.theia.antrea.io/v1alpha1
kind: NetworkPolicyRecommendation
metadata:
  name: pr-test
spec:
  jobType: "initial"
  policyType: "anp-deny-applied"
  executorInstances: 1
  driverCoreRequest: 200m
  driverMemory: 512M
  executorCoreRequest: 200m
  executorMemory: 512M
  excludeLabels: true
  toServices: true
```

You can observe the Spark Application starts, runs and stops. You could run the following command to get the details about the NetworkPolicyRecommendation anytime.

```bash
kubectl get npr pr-test -n flow-visibility -o json
```

The output would look like the following. When the job is finished, you will also be able to see the recommendedNetworkPolicy in the NetworkPolicyRecommendation status.

```bash
{
    "apiVersion": "crd.theia.antrea.io/v1alpha1",
    "kind": "NetworkPolicyRecommendation",
    ...
    "status": {
        "completedStages": 4,
        "sparkApplication": "97186a4d-5681-470d-82d6-e890ba539c34",
        "state": "RUNNING",
        "totalStages": 5,
    }
}

```

After deleting the NetworkPolicyRecommendation `pr-test`, the corresponding RecommendedNetworkPolicy CR and the row in ClickHouse database will be deleted at the same time.


pkg/apis/crd/v1alpha1/types.go
```diff
@@ -51,7 +59,12 @@ type NetworkPolicyRecommendationSpec struct {
 }
 
 type NetworkPolicyRecommendationStatus struct {
```
yuntanghsu, line 61: Thanks Yanjun for working this. 
Do you think we can add the start and termination time in Status as well? I think we might need these two information for CLI side?

yanjunz97: I notice the list command in CLI will display a completion time. Currently this timestamp is retrieved from the column `timeCreated` in the ClickHouse. I have created a `timeCreated` field in the `RecommendedNetworkPolicy` when reading from the ClickHouse. Do you still need another start/end time of the `NetworkPolicyRecommendation` for CLI?

yuntanghsu: I think we might need creationTime as well?
`sparkApplicationTable := [][]string{{"CreationTime", "CompletionTime", "ID", "Status"}}`

yanjunz97: I see. Let me add them.

pkg/apis/crd/v1alpha1/types.go
```diff
@@ -61,3 +76,28 @@ type NetworkPolicyRecommendationList struct {
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec RecommendedNetworkPolicySpec `json:"spec,omitempty"`
+}
+
+type RecommendedNetworkPolicySpec struct {
+	Id          string      `json:"id,omitempty"`
+	Type        string      `json:"resultType,omitempty"`
+	TimeCreated metav1.Time `json:"timeCreated,omitempty"`
```
yuntanghsu, line 93: I think RecommendedNetworkPolicySpec.TimeCreated and NetworkPolicyRecommendationStatus.EndTime are the same in Theia v0.2.0?
Do you think we need two similar fields here to track the end time? (No strong opinion, just curious when TimeCreated will be used.)

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -118,19 +199,64 @@ func (c *NPRecommendationController) Run(stopCh <-chan struct{}) {
...
+	if quit {
+		return false
+	}
+	defer c.deletionQueue.Done(obj)
+	if key, ok := obj.(NamespacedId); !ok {
+		c.queue.Forget(obj)
+		klog.ErrorS(nil, "Expected Spark Application namespaced id in work queue", "got", obj)
+		return true
+	} else if err := c.cleanupNPRecommendation(key.Namespace, key.Id); err == nil {
+		// If no error occurs we Forget this item so it does not get queued again until
```
yuntanghsu, line 230:
```suggestion
		// If no error occurs we forget this item so it does not get queued again until
```

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -139,7 +265,7 @@ func (c *NPRecommendationController) processNextWorkItem() bool {
 	defer c.queue.Done(obj)
 	if key, ok := obj.(apimachinerytypes.NamespacedName); !ok {
 		c.queue.Forget(obj)
-		klog.Errorf("Expected NP Recommendation in work queue but got %#v", obj)
+		klog.ErrorS(nil, "Expected NP Recommendation in work queue", "got", obj)
```
yuntanghsu, line 268:
```suggestion
		klog.ErrorS(nil, "Expected NP Recommendation not in work queue", "got", obj)
```

yanjunz97: I think this message means we expected NP Recommendation in work queue but go another object.

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: Fixed, thanks!

pkg/apis/crd/v1alpha1/types.go
yanjunz97: In Theia v0.2.0 we use `TimeCreated` as the end time for those jobs if we do not find the corresponding SparkApplication but find the result in the ClickHouse. But in this release, I think we can always use the `NetworkPolicyRecommendation.Status.EndTime` for the end time.

I keep this field in RecommendedNetworkPolicy just for the consistency with the ClickHouse row. I also do not have a strong preference on whether to remove it.

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -39,44 +50,106 @@ const (
 	maxRetryDelay = 300 * time.Second
 	// Default number of workers processing an Service change.
 	defaultWorkers = 4
+	// Time format for parsing input time
+	inputTimeFormat = "2006-01-02 15:04:05"
+	// Time format for parsing time from ClickHouse
+	clickHouseTimeFormat = "2006-01-02T15:04:05Z"
+	// SparkApplication id idex name for RecommendedNetworkPolicy
```
yuntanghsu, line 57:
```suggestion
	// SparkApplication id index name for RecommendedNetworkPolicy
```

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +294,418 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+		return fmt.Errorf("failed to delete the SparkApplication %s, error: %v", sparkApplicationId, err)
+	}
+	// Delete the RNP if exists
+	rnps, err := c.recommendedNPInformer.GetIndexer().ByIndex(idIndex, sparkApplicationId)
+	if err != nil {
+		return fmt.Errorf("failed to get the RecommendedNetworkPolicy when deleting, error: %v", err)
+	}
+	for _, obj := range rnps {
+		recommendedNetworkPolicy := obj.(*crdv1alpha1.RecommendedNetworkPolicy)
+		c.crdClient.CrdV1alpha1().RecommendedNetworkPolicies(namespace).Delete(context.TODO(), recommendedNetworkPolicy.Name, metav1.DeleteOptions{})
```
yuntanghsu, line 325: Unhandled error?

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +294,418 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+			if err != nil {
+				return err
+			}
+			// Create RecommendedNetworkPolicy CR
+			recommendedNetworkPolicy, err = c.crdClient.CrdV1alpha1().RecommendedNetworkPolicies(npReco.Namespace).Create(context.TODO(), recommendedNetworkPolicy, metav1.CreateOptions{})
+			if err != nil {
+				return err
+			}
+		}
+		klog.V(2).InfoS("Created RecommendedNetworkPolicy", "RecommendedNetworkPolicy", recommendedNetworkPolicy.Name, "NetworkRecommendationPolicy", npReco.Name)
```
yuntanghsu, line 378: Should this log be inside the "else" above?

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: Fixed, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -39,44 +50,106 @@ const (
...
+func idIndexFunc(obj interface{}) ([]string, error) {
+	recommendedNP, ok := obj.(*crdv1alpha1.RecommendedNetworkPolicy)
+	if !ok {
+		return nil, fmt.Errorf("obj is not RecommendedNetworkPolicy: %v", obj)
+	}
+	return []string{recommendedNP.Spec.Id}, nil
+}
+
 func (c *NPRecommendationController) addNPRecommendation(obj interface{}) {
 	npReco, _ := obj.(*crdv1alpha1.NetworkPolicyRecommendation)
```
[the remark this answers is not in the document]
yanjunz97: Added, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: Good catch, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +300,420 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+	if err == nil {
+		c.addPeriodicSync(apimachinerytypes.NamespacedName{
+			Name:      npReco.Name,
+			Namespace: npReco.Namespace,
+		})
+	}
+	return err
+}
+
+func (c *NPRecommendationController) startSparkApplication(npReco *crdv1alpha1.NetworkPolicyRecommendation) error {
```
wsquan171, line 501: can we abstract these validation logic to some querrier-like interface so it can be shared with API server REST handler? This way we can move the error reporting at REST CREATE time. Otherwise a success response from API server could end up in invalid state due to user input error, which looks confusing.  cc @yuntanghsu

wsquan171: As Yanjun pointed out offline, as the similar checks are also done on CLI, and no other component / user is expected to be directly calling the API server at the moment, validation at REST handler doesn't provide much benefit at the moment. probably can skip this at the moment until when UI is to be added and we can reevaluate.

yuntanghsu: Will the user directly create a NPR CR to start the SparkApplication? If no, maybe we can only validate the input on CLI side.

yanjunz97: We do not suggest user to do this. But we do not prevent user from doing this. I think the validation here could be a protection in case this happens.

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -118,19 +205,64 @@ func (c *NPRecommendationController) Run(stopCh <-chan struct{}) {
...
+
+	go wait.Until(c.deletionworker, time.Second, stopCh)
+
 	for i := 0; i < defaultWorkers; i++ {
 		go wait.Until(c.worker, time.Second, stopCh)
 	}
 	<-stopCh
 }
 
+func (c *NPRecommendationController) deletionworker() {
```
wsquan171, line 220: We need some GC logic to recycle stale resources (rnp and db entry) in case npr is removed during manager downtime, or manager is restarted without finishing the cleanup job. this may be addressed in a future pr.

pkg/controller/networkpolicyrecommendation/util.go
```diff
@@ -0,0 +1,326 @@
...
+	url := fmt.Sprintf("%s?debug=false&username=%s&password=%s", endpoint, username, password)
+	connect, err = connectClickHouse(url)
+	if err != nil {
+		return nil, fmt.Errorf("error when connecting to ClickHouse, %v", err)
+	}
+	return connect, nil
+}
+
+func getPolicyRecommendationResult(client kubernetes.Interface, id string, namespace string) (recommendedNetworkPolicy *crdv1alpha1.RecommendedNetworkPolicy, err error) {
+	connect, err := setupClickHouseConnection(client, namespace)
```
wsquan171, line 241: can we cache / resue connection within the controller instead of reestablishing connection each time db check is needed?

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +300,420 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+	// Check the status before checking the progress in case the job is failed or completed
+	state, err := c.checkSparkApplicationStatus(npReco)
+	if err != nil {
+		return err
+	}
+	if state != crdv1alpha1.NPRecommendationStateRunning {
+		return nil
+	}
+	service := fmt.Sprintf("pr-%s-ui-svc", npReco.Status.SparkApplication)
+	serviceIP, servicePort, err := getServiceAddr(c.kubeClient, service, npReco.Namespace, sparkPortName)
```
wsquan171, line 407: feels like it's a bit wasteful given the service addr won't change, yet multiple api calls to k8s apiserver is done each time we do periodical sync. Also given that manager is now running in k8s cluster (in hence have access to coreDNS), can we rather access the spark svc via DNS name?

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -39,44 +50,112 @@ const (
...
 		resyncPeriod,
 	)
 
+	// Add SparkApplication ID index for RecommendedNetworkPolicy
+	recommendedNPInformer.Informer().AddIndexers(cache.Indexers{idIndex: idIndexFunc})
+
 	return c
 }
 
+func idIndexFunc(obj interface{}) ([]string, error) {
```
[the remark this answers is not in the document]
yanjunz97: Rename it to `rnpIdIndexFunc`.

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -39,44 +50,112 @@ const (
...
+	}
+	return []string{recommendedNP.Spec.Id}, nil
+}
+
 func (c *NPRecommendationController) addNPRecommendation(obj interface{}) {
-	npReco, _ := obj.(*crdv1alpha1.NetworkPolicyRecommendation)
-	klog.V(2).Infof("Processing NP Recommendation %s ADD event, labels: %v", npReco.Name, npReco.Labels)
+	npReco, ok := obj.(*crdv1alpha1.NetworkPolicyRecommendation)
+	if !ok {
+		klog.V(2).InfoS("fail to convert to NetworkPolicyRecommendation", "object", obj)
```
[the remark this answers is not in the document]
yanjunz97: Updated, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +300,420 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+}
+
+func (c *NPRecommendationController) startJob(npReco *crdv1alpha1.NetworkPolicyRecommendation) error {
+	// Validate Cluster readiness
+	if err := validateCluster(c.kubeClient, npReco.Namespace); err != nil {
+		return err
+	}
+	err := c.startSparkApplication(npReco)
+	// Mark the NetworkPolicyRecommendation as failed and not retry if it failed due to invalid request
+	if err != nil && strings.Contains(err.Error(), "invalid request") {
```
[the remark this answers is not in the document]
yanjunz97: Sounds great. Updated, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +300,420 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+	// Delete the RNP if exists
+	rnps, err := c.recommendedNPInformer.GetIndexer().ByIndex(idIndex, sparkApplicationId)
+	if err != nil {
+		return fmt.Errorf("failed to get the RecommendedNetworkPolicy when deleting, error: %v", err)
+	}
+	for _, obj := range rnps {
+		recommendedNetworkPolicy := obj.(*crdv1alpha1.RecommendedNetworkPolicy)
+		err := c.crdClient.CrdV1alpha1().RecommendedNetworkPolicies(namespace).Delete(context.TODO(), recommendedNetworkPolicy.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to delete the RecommendedNetworkPolicy %s, error: %v", recommendedNetworkPolicy.Name, err)
```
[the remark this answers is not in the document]
yanjunz97: Updated, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +300,420 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+		nsAllowListStr := "["
+		for i, nsAllow := range npReco.Spec.NSAllowList {
+			nsAllowListStr += ("\"" + nsAllow + "\"")
+			if i != len(npReco.Spec.NSAllowList)-1 {
+				nsAllowListStr += ","
+			}
+		}
+		nsAllowListStr += "]"
+		recoJobArgs = append(recoJobArgs, "--ns_allow_list", nsAllowListStr)
+	}
```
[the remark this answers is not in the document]
yanjunz97: Updated, thanks!

pkg/controller/networkpolicyrecommendation/util.go
yanjunz97: Updated, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: Updated, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: The idea of GC sounds great to me. As we may remove rnp in next release, maybe it will be a periodic job checking db to remove an entry with an id not in the current npr list?
Besides, I'm also thinking about whether to do the similar GC for the stale SparkApplication. I'm a little bit worry about whether this will delete some jobs not created by the manager in the future.

pkg/controller/networkpolicyrecommendation/controller.go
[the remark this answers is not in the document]
wsquan171: still think we should return in case of conversion to NetworkPolicyRecommendation failed. Otherwise panic will be triggered.

pkg/controller/networkpolicyrecommendation/controller.go
wsquan171: one way around is we introduce additional metadata (annotation / labels) for the ones created by Theia. This way we won't bother with other unrelated SparkApplications. Periodical is good but at least this should be triggered once per mgr init.

wsquan171 (commented):
overall lgtm. will check again after rebase and approve.

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -168,11 +306,425 @@ func (c *NPRecommendationController) syncNPRecommendation(key apimachinerytypes.
...
+	rnps, err := c.recommendedNPInformer.GetIndexer().ByIndex(idIndex, sparkApplicationId)
+	if err != nil {
+		return fmt.Errorf("failed to get the RecommendedNetworkPolicy when deleting, error: %v", err)
+	}
+	var aggregatedErrorMsg string
+	for _, obj := range rnps {
+		recommendedNetworkPolicy := obj.(*crdv1alpha1.RecommendedNetworkPolicy)
+		err = c.crdClient.CrdV1alpha1().RecommendedNetworkPolicies(namespace).Delete(context.TODO(), recommendedNetworkPolicy.Name, metav1.DeleteOptions{})
+		if err != nil {
+			aggregatedErrorMsg += fmt.Sprintf("failed to delete the RecommendedNetworkPolicy %s, error: %v\n", recommendedNetworkPolicy.Name, err)
```
wsquan171, line 340: typically we don't want to have multi-line error messages. could collect failed name and err in 2 lists and call fmt when returning error.

yanjunz97: Make sense to me. Updated, thanks!

yuntanghsu (approved):
LGTM

wsquan171 (approved):
LGTM

PR #116, opened 2022-09-26
Support setting securityContext & Fixed existing zookeper 

ColonelBundy:


build/charts/theia/values.yaml
```diff
@@ -109,6 +109,9 @@ clickhouse:
         repository: "projects.registry.vmware.com/antrea/theia-zookeeper"
         pullPolicy: "IfNotPresent"
         tag: "3.8.0"
+      # Enable securitycontextconstraints
+      # set to false for openshift
+      scc: true
```
yanjunz97, line 114: I think the name `scc` and the comments are not clear enough. I would suggest some updates like the follows. Please keep the `--` in the comments so that it can be included to when generating the documents. 

```suggestion
      # -- Determine whether to run ZooKeeper as a non-privileged user.
      # Please set this value to false when running on OpenShift.
      securityContext:
        enable: true
```

yanjunz97 (commented):
Thanks @ColonelBundy for contributing!
I add some comments. Could you also run `make` under the `build/charts` to update the documents to pass our check on verifying doc. Thank you!

build/charts/theia/values.yaml
ColonelBundy: Instead of making this Openshift specific I believe it's better to make the securityContext configurable.

ColonelBundy:
> Thanks @ColonelBundy for contributing! I add some comments. Could you also
> 
> * Run `make` under the `build/charts` to update the documents to pass our check on verifying doc.
> * Sign your commit by using `git commit -s` when committing to pass our DCO check.
> 
> Thank you!

Thanks for the feedback, I made some updates.

build/charts/theia/values.yaml
```diff
@@ -109,6 +109,11 @@ clickhouse:
         repository: "projects.registry.vmware.com/antrea/theia-zookeeper"
         pullPolicy: "IfNotPresent"
         tag: "3.8.0"
+      # -- Set securityContext
+      # Use a specific uid, gid for zookeeper
```
yanjunz97, line 113: Could you add `.` after each sentence? And also re-generate the doc.

build/charts/theia/values.yaml
```diff
@@ -121,6 +126,12 @@ grafana:
   # -- Determine whether to install Grafana. It is used as a data visualization
   # and monitoring tool.  
   enable: true
+  # -- Set securityContext
+  # Use a specific uid, gid for grafana
```
yanjunz97, line 130: Same here

yanjunz97 (commented):
LGTM, only a nit.

build/charts/theia/values.yaml
ColonelBundy: Fixed.

build/charts/theia/values.yaml
ColonelBundy: Fixed.

yanjunz97 (approved):

PR #118, opened 2022-09-27
Update Grafana Panel Plugins

heanlan:
This PR does the following two things:

When testing with snowflake datasource plugin, we found it will
automatically convert all fields name to upper case, which will
lead to string matching failure in our sankey-panel plugin and
chord-panel plugin, as we are doing case sensitive string matching.
As a work around, we do the following change to both sankey and
chord plugins: when search for fields with names in data frame,
do case insenstive string matching.

When testing with snowflake datasource plugin, we found for a field
with type String, when the value is Null, it will not convert a
Null to ''. So we change our chord panel plugin to also check for
Null value, to be compatible with different datasources.


dreamtalen (approved):

yuntanghsu (approved):
LGTM

heanlan:
@dreamtalen @yuntanghsu thanks for the quick review.

I just merged my other changes of plugins in #117 into this PR. I previously came up with a "workaround" on Grafana dashboard, with which I thought the changes are no longer needed, so I closed that PR. Just found the workaround still not work, so I added the changes back. Please help take another look when you have time. Still a small PR.

yanjunz97 (approved):

PR #121, opened 2022-10-07
Fix bugs related to ClickHouse data schema upgrading

yanjunz97:
This PR fixes some bugs regarding ClickHouse data schema upgrading

- Set TTL in a separate command instead of in table creation command to avoid TTL not being updated when table is already created
- Fixed some bugs in data schema management to make sure upgrading from v0.1.0 to v0.3.0 is correct


yuntanghsu:
Do we need to change ttl when downgrading to v0.1.0?

yanjunz97:
> Do we need to change ttl when downgrading to v0.1.0? Is it because we only have table flows in v0.1.0?

We do not support downgrading to v0.1.0 as the data schema management is introduced in v0.2.0.

yuntanghsu (approved):
LGTM

dreamtalen (approved):
LGTM

PR #123, opened 2022-10-13
Add unit tests for Snowflake backend

heanlan:
This PR adds some initial unit tests for Snowflake backend code, and adds a github workflow to run the tests. It also fixes code coverage reporting for unit tests following:
https://github.com/antrea-io/antrea/pull/4172


antoninbas (commented):
I am not sure why you are building fake implementations for the S3 clients instead of just mocking the interfaces and using expectations in tests?

heanlan:
> I am not sure why you are building fake implementations for the S3 clients instead of just mocking the interfaces and using expectations in tests?

I was just thinking both way would work. Should we mock the interface? I can switch to mock. Dumb question: actually I'm not sure when should we mock the interface and when should we create fake implementation.

antoninbas:
> > I am not sure why you are building fake implementations for the S3 clients instead of just mocking the interfaces and using expectations in tests?
> 
> I was just thinking both way would work. Should we mock the interface? I can switch to mock for both S3 and sqs client. Dumb question: actually I'm not sure when should we mock the interface and when should we create fake implementation.

I would say go with mocks when possible:
- less work in general
- more readable tests
- can validate arguments

Even though you can do it with mocks as well, a fake implementation can make sense if you need a lot of side effects (actions you need to take when the method is called by your tests).

If you need to add some methods which will be called by the test cases, it can also make sense to have a fake implementation. One example is if you want to abstract a "clock" behind an interface.

Example:
```go
type object struct {
    c clock.Interface
    client api.Interface
}

func NewObject(c clock.Interface, client api.Interface) *object {}
```

`object` is the object you want to write unit tests for. You probably want to use a fake clock for `c` (in order to have a virtual clock that you can manually advance) and a mock for the API `client`.

snowflake/cmd/createBucket_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	origLogger := logger
+	defer func() {
+		logger = origLogger
+	}()
+
+	var b bytes.Buffer
+	logger = sftesting.NewLogger(&b)
+
+	for _, tc := range []struct {
```
antoninbas, line 44: in a case like this one where you are only going to gave 2 cases (new bucket / existing bucket) and where most of the test case code ends up being conditioned by `tc.bucketExist`, I would recommend _not_ using table-driven tests. You can just write:
```go
t.Run("Create existing bucket", func(t *testing.T) {
// ...
})
t.Run("Create new bucket", func(t *testing.T) {
// ...
})
```
It's more readable and not really more verbose

snowflake/cmd/createKmsKey_test.go
```diff
@@ -0,0 +1,64 @@
...
+const (
+	creatingKeyLog = "Creating key"
+	createdKeyLog  = "Created key"
+)
+
+func TestCreateKey(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	origLogger := logger
```
antoninbas, line 39: I don't think it is very useful to validate the log messages in this test

I think this comment may apply to the other tests as well: IMO you want to rely on log verification when you have no other way to determine whether the correct code path is being executed. But in our case, I feel like we can always check the code path by checking the mock methods being called. Still, you can decide whether we should keep these checks or not.

snowflake/cmd/deleteBucket_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	origLogger := logger
+	defer func() {
+		logger = origLogger
+	}()
+
+	var b bytes.Buffer
+	logger = sftesting.NewLogger(&b)
+
+	for _, tc := range []struct {
```
antoninbas, line 50: same comment as above regarding table-driven tests

snowflake/cmd/receiveSqsMessage.go
```diff
@@ -97,7 +99,7 @@ func receiveSQSMessage(ctx context.Context, sqsClient sqsclient.Interface, queue
 		return nil
 	}
 	message := output.Messages[0]
-	fmt.Println(*message.Body)
+	println(*message.Body)
```
antoninbas, line 102: rather than this change, I'd prefer to add an extra parameter to the `receiveSQSMessage` function (of type `io.Writer`). It's cleaner and a very common solution.

snowflake/pkg/infra/manager_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"path/filepath"
+	"testing"
+
+	"antrea.io/theia/snowflake/database"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWriteMigrationsToDisk(t *testing.T) {
+	workdir, err := os.Getwd()
```
antoninbas, line 28: you should create a temporary directory and use it, instead of using the working directory

snowflake/pkg/infra/manager_test.go
```diff
@@ -0,0 +1,40 @@
...
+	require.NoError(t, err)
+	err = writeMigrationsToDisk(database.Migrations, database.MigrationsPath, filepath.Join(workdir, migrationsDir))
+	require.NoError(t, err)
+	entries, err := database.Migrations.ReadDir(database.MigrationsPath)
+	require.NoError(t, err)
+	for _, entry := range entries {
+		_, err := os.Stat(filepath.Join(workdir, migrationsDir, entry.Name()))
+		assert.NoErrorf(t, err, "Migration file %s not exist", entry.Name())
+	}
+	err = os.RemoveAll(filepath.Join(workdir, migrationsDir))
```
antoninbas, line 38: should be done with a `defer` statement

snowflake/pkg/snowflake/dsn.go
```diff
@@ -47,7 +49,7 @@ func GetDSN(options ...func(*sf.Config)) (string, *sf.Config, error) {
 			return value
 		}
 		if failOnMissing {
-			log.Fatalf("%v environment variable is not set.", k)
+			logFatalf("%v environment variable is not set.", k)
```
antoninbas, line 52: I think we should just change that logic and have the `GetDSN` function return an error instead of calling `log.Fatal`

snowflake/pkg/snowflake/dsn_test.go
```diff
@@ -0,0 +1,118 @@
...
+				assert.Equal(t, tc.expectedDSN, dsn)
+				if tc.expectedCFG != nil {
+					assert.Equal(t, tc.expectedCFG.Account, cfg.Account)
+					assert.Equal(t, tc.expectedCFG.User, cfg.User)
+					assert.Equal(t, tc.expectedCFG.Password, cfg.Password)
+					assert.Equal(t, tc.expectedCFG.Host, cfg.Host)
+					assert.Equal(t, tc.expectedCFG.Port, cfg.Port)
+					assert.Equal(t, tc.expectedCFG.Protocol, cfg.Protocol)
+				}
+				assert.True(t, errors.Is(err, tc.expectedError))
```
antoninbas, line 114: there are some test assertions for errors (e.g., `assert.ErrorIs`), you don't need to use `assert.True`

snowflake/pkg/snowflake/snowflake_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"go.uber.org/zap"
+)
+
+const warehouseName = "ANTREA-QUERIES"
+
+func TestCreateWarehouse(t *testing.T) {
+	db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
+	if err != nil {
+		t.Fatalf("an error '%s' was not expected when opening a stub database connection", err)
+	}
```
antoninbas, line 35: let's stick to using `require.NoError` for uniformity (plus it's less verbose)

snowflake/Makefile
```diff
@@ -7,9 +7,17 @@ all: bin
 bin:
```
antoninbas, line 7: we need a `generate` Make target so that we can easily re-generate the mocks

see https://betterprogramming.pub/a-gomock-quick-start-guide-71bee4b3a6f1 for a reference

heanlan: thanks, addressed all the comments above.

I'm not confident on the generate gomock part, feeling I'm doing it in a strange way. Could you help take a look?

snowflake/pkg/snowflake/snowflake_test.go
```diff
@@ -0,0 +1,186 @@
...
+			execErr := c.DropWarehouse(context.TODO(), warehouseName)
+			if err := mock.ExpectationsWereMet(); err != nil {
+				t.Errorf("there were unfulfilled expectations: %s", err)
+			}
+			assert.Equal(t, tc.expectedError, execErr)
+		})
+	}
+}
+
+func initializeLogger() (logr.Logger, error) {
```
antoninbas, line 179: you should use https://pkg.go.dev/github.com/go-logr/logr/testing

snowflake/pkg/snowflake/snowflake_test.go
```diff
@@ -0,0 +1,186 @@
...
+const warehouseName = "ANTREA-QUERIES"
+
+func TestCreateWarehouse(t *testing.T) {
+	db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
+	require.NoErrorf(t, err, "an error was not expected when opening a stub database connection")
+	defer db.Close()
+
+	logger, err := initializeLogger()
+	if err != nil {
+		t.Fatalf("error while initializing logger: %s", err)
```
antoninbas, line 39: please use `require.NoError` consistently

snowflake/pkg/snowflake/snowflake_test.go
```diff
@@ -0,0 +1,186 @@
...
+			},
+			expectedError: fmt.Errorf("some error"),
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.prepareMock(mock)
+			execErr := c.CreateWarehouse(context.TODO(), warehouseName, config)
+			if err := mock.ExpectationsWereMet(); err != nil {
+				t.Errorf("there were unfulfilled expectations: %s", err)
+			}
```
antoninbas, line 83: please use `assert.NoError`

heanlan: thanks for the comments. addressed them all.

snowflake/Makefile
```diff
@@ -7,9 +7,22 @@ all: bin
...
+test-unit: .coverage
+	@echo
+	@echo "==> Running unit tests <=="
+	$(GO) test -race -coverpkg=antrea.io/theia/snowflake/cmd/...,antrea.io/theia/snowflake/pkg/... \
+	  -coverprofile=.coverage/coverage-unit.txt -covermode=atomic \
+	  antrea.io/theia/snowflake/cmd/... antrea.io/theia/snowflake/pkg/... 
+
+.PHONY: generate
+generate:
+	MOCKGEN_COPYRIGHT_FILE=$(PWD)/hack/boilerplate/license_header.raw.txt \
```
antoninbas, line 24: please use `$(CURDIR)` like we do in other places in this Makefile

https://stackoverflow.com/questions/52437728/bash-what-is-the-difference-between-pwd-and-curdir

snowflake/pkg/infra/manager_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestWriteMigrationsToDisk(t *testing.T) {
+	tempDir, err := os.MkdirTemp("", "antrea-pulumi-test")
+	require.NoError(t, err)
+	defer func() {
+		err = os.RemoveAll(filepath.Join(tempDir, migrationsDir))
+		require.NoError(t, err)
+	}()
```
antoninbas, line 33: just `defer os.RemoveAll(tempDir)`

antoninbas (commented):
couple more nits, otherwise LGTM

snowflake/Makefile
heanlan: thanks, addressed

antoninbas (approved):

snowflake/pkg/snowflake/snowflake_test.go
```diff
@@ -0,0 +1,159 @@
...
+	intiallySuspended := true
+	config := WarehouseConfig{
+		Size:               &warehouseSize,
+		MinClusterCount:    &minClusterCount,
+		MaxClusterCount:    &maxClusterCount,
+		ScalingPolicy:      &scalingPolicy,
+		AutoSuspend:        &autoSuspend,
+		InitiallySuspended: &intiallySuspended,
+	}
+	query := fmt.Sprintf("CREATE WAREHOUSE %s WITH WAREHOUSE_SIZE = XSMALL MIN_CLUSTER_COUNT = 1 MAX_CLUSTER_COUNT = 3 SCALING_POLICY = STANDARD AUTO_SUSPEND = 60 INITIALLY_SUSPENDED = true", warehouseName)
```
dreamtalen, line 51: Will it be better to format these variables `warehouseSize`, ..., `intiallySuspended` as well in this string?

heanlan: yes! thanks for catching it

dreamtalen (approved):
LGTM

PR #127, opened 2022-10-19
Enhance user experience

yuntanghsu:
This PR enhance the user experience when user use theia cli.
1. Increase the verbosity to prevent cli from printing not helpful message generated by util/env.go
2. When user use cli to use "policy-recommendation run" or "policy-recommendation retrieve" with --file option, we don't print the recommended result.


dreamtalen (approved):
LGTM

yanjunz97 (approved):
LGTM

PR #132, opened 2022-10-24
Modify clickhouse statistic related func for theia-manager

yuntanghsu:
This PR modifies and adds the clickhouse statistic related functions for Theia Manager:

1. Define REST request and return payload struct for related API endpoints and add code generation under pkg/apis/stats
    a. REST endpoint apis/stats.theia.antrea.io/v1alpha1/clickhouse
2. Adapt CLI to use Theia Manager
3. Modify e2e test

The changes are split into two chunks (to hopefully help make reviews easier), with one commit focusing on adding status API, another one implementing the CLI changes and modifying the e2e test,


pkg/util/utils.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/ClickHouse/clickhouse-go"
+	"github.com/google/uuid"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var OpenSql = sql.Open
```
wsquan171, line 29:
```suggestion
var SqlOpenFunc = sql.Open
```

pkg/controller/stats/controller.go
```diff
@@ -0,0 +1,227 @@
...
+	shardNum() as Shard,
+	arrayStringConcat(arrayMap(x -> demangle(addressToSymbol(x)), trace), '\\n') AS trace_function,
+	count()
+FROM cluster('{cluster}', system.stack_trace)
+GROUP BY trace_function, Shard
+ORDER BY count()
+DESC SETTINGS allow_introspection_functions=1`,
+}
+
+type StatsController struct {
```
wsquan171, line 131: Naming of "controller" hints that there's some long-running routines involved, which is not the case here. I think we can simply call it ClickHouseStatQuerier or add "impl" suffix to distinguish from the interface. filename could also be renamed and mention clickhouse in case we'd want to collect stat from other components in the future.

pkg/controller/stats/controller.go
```diff
@@ -0,0 +1,227 @@
...
+	var err error
+	if c.clickhouseConnect == nil {
+		c.clickhouseConnect, err = util.SetupClickHouseConnection(c.kubeClient, namespace)
+		if err != nil {
+			return nil, err
+		}
+	}
+	result, err := c.clickhouseConnect.Query(queryMap[query])
+	if err != nil {
+		return nil, fmt.Errorf("failed to get data from clickhouse: %v", err)
```
wsquan171, line 187: do we want to reset db connection in case certain connection error is encountered, so that upon next query we can retry establishing a new conn?

pkg/apis/stats/v1alpha1/types.go
```diff
@@ -0,0 +1,28 @@
...
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ClickHouseStats struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Result [][]string `json:"jobType,omitempty"`
```
wsquan171, line 27: I think we should re-iterate `StatType` and rename `Result` to `Stat` to be clear. Also need to fix typo in the json tag.

wsquan171 (commented):
overall lgtm. some nits.

pkg/apis/stats/v1alpha1/types.go
```diff
@@ -0,0 +1,28 @@
...
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ClickHouseStats struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Stat [][]string `json:"Stat,omitempty"`
```
wsquan171, line 27:
```suggestion
	Stat [][]string `json:"stat,omitempty"`
```

wsquan171 (approved):
LGTM. one simple nit.

pkg/apis/stats/v1alpha1/types.go
yuntanghsu: Fixed. Thanks

pkg/controller/stats/clickhouse_stats.go
```diff
@@ -0,0 +1,228 @@
+// Copyright 2022 Antrea Authors.
```
yanjunz97, line 1: Just for curiosity, why this part is put in controller instead of apiserver pkg?

pkg/util/utils.go
```diff
@@ -0,0 +1,115 @@
...
+	var servicePort int
+	service, err := client.CoreV1().Services(serviceNamespace).Get(context.TODO(), serviceName, metav1.GetOptions{})
+	if err != nil {
+		return serviceIP, servicePort, fmt.Errorf("error when finding the Service %s: %v", serviceName, err)
+	}
+	serviceIP = service.Spec.ClusterIP
+	for _, port := range service.Spec.Ports {
+		if port.Name == servicePortName {
+			servicePort = int(port.Port)
+		}
```
yanjunz97, line 55: Maybe we can check `port.Protocol` as we do previously in CLI util instead of `port.Name` to avoid the naming problem?

pkg/util/utils.go
```diff
@@ -0,0 +1,115 @@
+// Copyright 2022 Antrea Authors.
```
yanjunz97, line 1: I also tried to move some utility functions to the common pkg in PR #138 . I think most of the functions are similar, but I would like to get some feedbacks about if we prefer to put all these functions in one file like this one, or classify them into multiple pkgs like clickhouse, policyrecommendation, k8s as I do in PR #138

pkg/apis/stats/v1alpha1/types.go
```diff
@@ -0,0 +1,28 @@
...
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ClickHouseStats struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Stat [][]string `json:"stat,omitempty"`
```
yanjunz97, line 27: Maybe we could make this as a structured data type instead of a matrix of string? Like
```golang
type Stats struct {
 Diskinfos []diskinfo `json:"diskinfos,omitempty"`
 ...
}
```
I just feel it is unclear what kind of Stat would be included when looking at this API.

pkg/controller/stats/clickhouse_stats.go
yuntanghsu: Thanks for the reviewing. It shouldn't be there as it doesn't contain functionality of controller. I've moved it to apiserver/utils/stats.

pkg/util/utils.go
yuntanghsu: Sure. Updated.

pkg/util/utils.go
yuntanghsu: I don't have strong opinion regarding this, but I might prefer classifying them into multiple pkgs.

pkg/apis/stats/v1alpha1/types.go
yuntanghsu: Updated. Please take a look to see if it is clear now?

build/charts/theia/templates/theia-manager/service.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
 spec:
   ports:
     - port: {{ .Values.theiaManager.apiServer.apiPort }}
+      name: tcp
```
yanjunz97, line 12: Maybe we do not need this now?

pkg/apis/stats/v1alpha1/types.go
```diff
@@ -0,0 +1,63 @@
...
+	StackTraces []StackTrace `json:"stackTraces,omitempty"`
+}
+
+type DiskInfo struct {
+	Shard          string `json:"shard,omitempty"`
+	Database       string `json:"name,omitempty"`
+	Path           string `json:"path,omitempty"`
+	FreeSpace      string `json:"freeSpace,omitempty"`
+	TotalSpace     string `json:"totalSpace,omitempty"`
+	UsedPercentage string `usedPercentage:"name,omitempty"`
```
yanjunz97, line 41:
```suggestion
	UsedPercentage string `json:"usedPercentage,omitempty"`
```

build/charts/theia/templates/theia-manager/service.yaml
yuntanghsu: Removed. Thanks.

pkg/apis/stats/v1alpha1/types.go
yuntanghsu: Fixed.

ci/jenkins/test-vmc.sh
```diff
@@ -339,7 +339,7 @@ function deliver_antrea {
 
     ${GIT_CHECKOUT_DIR}/hack/generate-manifest.sh --ch-size 100Mi --ch-monitor-threshold 0.1 > ${GIT_CHECKOUT_DIR}/build/yamls/flow-visibility.yml
     ${GIT_CHECKOUT_DIR}/hack/generate-manifest.sh --no-grafana --spark-operator --theia-manager > ${GIT_CHECKOUT_DIR}/build/yamls/flow-visibility-with-spark.yml
-    ${GIT_CHECKOUT_DIR}/hack/generate-manifest.sh --no-grafana > ${GIT_CHECKOUT_DIR}/build/yamls/flow-visibility-ch-only.yml
+    ${GIT_CHECKOUT_DIR}/hack/generate-manifest.sh --no-grafana --theia-manager > ${GIT_CHECKOUT_DIR}/build/yamls/flow-visibility-ch-only.yml
```
salv-orlando, line 342: were we lacking theia manager in the manifests we use for testing before this change?

pkg/apiserver/utils/stats/clickhouse_stats.go
```diff
@@ -0,0 +1,195 @@
...
+			res := v1alpha1.DiskInfo{}
+			err = result.Scan(&res.Shard, &res.Database, &res.Path, &res.FreeSpace, &res.TotalSpace, &res.UsedPercentage)
+			res.UsedPercentage = res.UsedPercentage + " %"
+			stats.DiskInfos = append(stats.DiskInfos, res)
+		case tableInfoQuery:
+			res := v1alpha1.TableInfo{}
+			var totalRows sql.NullString
+			var totalBytes sql.NullString
+			err = result.Scan(&res.Shard, &res.Database, &res.TableName, &totalRows, &totalBytes, &res.TotalCols)
+			if !totalRows.Valid || !totalBytes.Valid {
```
salv-orlando, line 175: Is it worth checking err!=nil before this line?

pkg/apiserver/utils/stats/clickhouse_stats.go
```diff
@@ -0,0 +1,195 @@
...
+		return fmt.Errorf("failed to get data from clickhouse: %v", err)
+	}
+	defer result.Close()
+	for result.Next() {
+		var err error
+		switch query {
+		case diskQuery:
+			res := v1alpha1.DiskInfo{}
+			err = result.Scan(&res.Shard, &res.Database, &res.Path, &res.FreeSpace, &res.TotalSpace, &res.UsedPercentage)
+			res.UsedPercentage = res.UsedPercentage + " %"
```
salv-orlando, line 168: Is it worth checking err!=nil before this line?

pkg/apiserver/utils/stats/clickhouse_stats.go
```diff
@@ -0,0 +1,195 @@
...
+			res := v1alpha1.InsertRate{}
+			err = result.Scan(&res.Shard, &res.RowsPerSec, &res.BytesPerSec)
+			stats.InsertRates = append(stats.InsertRates, res)
+		case stackTraceQuery:
+			res := v1alpha1.StackTrace{}
+			err = result.Scan(&res.Shard, &res.TraceFunctions, &res.Count)
+			stats.StackTraces = append(stats.StackTraces, res)
+		}
+		if err != nil {
+			return fmt.Errorf("failed to parse the data returned by database: %v", err)
```
salv-orlando, line 191: This is just to make a little less ignorant... can this query return multiple rows?

In that case, would it better to skip and move to the next record if we fail to parse a record?

salv-orlando (commented):
Overall LGTM, just some nits inline

ci/jenkins/test-vmc.sh
yuntanghsu: We previously added theia manager at line #341 for NPR related functions in CLI. Line #342 is for the clickhouse related changes in this PR.

pkg/apiserver/utils/stats/clickhouse_stats.go
yuntanghsu: The query returns multiple rows.
I think it is a good idea to skip the error row instead of throwing away the whole data?
cc @yanjunz97 and @wsquan171 for some feedbacks.

yanjunz97: Agree to skip the error row. It would be better if we could display the results retrieved successfully and the error regarding to the error row at the same time?

pkg/apiserver/utils/stats/clickhouse_stats.go
yuntanghsu: Fixed. Thanks.

pkg/apiserver/utils/stats/clickhouse_stats.go
yuntanghsu: Fixed. Thanks.

pkg/apiserver/utils/stats/clickhouse_stats.go
yuntanghsu: I added a ErrorMsg, which is a []string, in ClickHouseStats. If we have an error row, we will skip it and add the error msg to ErrorMsg. On CLI side, it will show both ErrorMsgs and the results.

wsquan171: The new approach looks good to me.

yanjunz97 (approved):
LGTM

PR #133, opened 2022-10-25
Fix the ingestion error of snowflake on Pod label fields

dreamtalen:
In this commit, we fix the ingestion error of snowpipe on flow records with Pod label fields. The reason behind this error is commas are used to separate different columns in CSV and Pod label columns contain commas so they are separated into multiple columns when ingested into snowflake.
To fix it, we add a customized file format for flow records and use single quote enclosing Pod label fields.

Changes on Antrea side: https://github.com/antrea-io/antrea/pull/4334

### DIFF
--- a/snowflake/pkg/infra/stack.go
+++ b/snowflake/pkg/infra/stack.go
@@ -292,7 +292,8 @@ func declareSnowflakeDatabase(
 			AutoIngest:       pulumi.Bool(true),
 			ErrorIntegration: notificationIntegration.ID(),
 			// FQN required for table and stage, see https://github.com/pulumi/pulumi-snowflake/issues/129
-			CopyStatement: pulumi.Sprintf("COPY INTO %s.%s.%s FROM @%s.%s.%s FILE_FORMAT = (TYPE = 'CSV')", databaseName, schemaName, flowsTableName, databaseName, schemaName, ingestionStageName),
+			// 0x27 is the hex representation of single quote. We use it to enclose Pod labels string.
+			CopyStatement: pulumi.Sprintf("COPY INTO %s.%s.%s FROM @%s.%s.%s FILE_FORMAT = (TYPE = CSV FIELD_OPTIONALLY_ENCLOSED_BY='0x27')", databaseName, schemaName, flowsTableName, databaseName, schemaName, ingestionStageName),
 		}, pulumi.Parent(schema), pulumi.DependsOn([]pulumi.Resource{ingestionStage, dbMigrations}), pulumi.DeleteBeforeReplace(true))
 		if err != nil {
 			return nil, err


heanlan (approved):
LGTM

snowflake/pkg/infra/stack.go
```diff
@@ -284,6 +284,19 @@ func declareSnowflakeDatabase(
...
 
+		// a bit of defensive programming: explicit dependency on ingestionStage may not be strictly required
+		fileFormat, err := snowflake.NewFileFormat(ctx, "antrea-sf-flows-format", &snowflake.FileFormatArgs{
+			Database:   db.ID(),
+			Schema:     schema.Name,
+			Name:       pulumi.String(flowFileFormatName),
+			FormatType: pulumi.String("CSV"),
+			// 0x27 is the hex representation of single quote. We use it to enclose Pod labels string.
+			FieldOptionallyEnclosedBy: pulumi.StringPtr("0x27"),
+		}, pulumi.Parent(schema), pulumi.DependsOn([]pulumi.Resource{ingestionStage, dbMigrations}), pulumi.DeleteBeforeReplace(true))
```
antoninbas, line 295: I don't think you need the `pulumi.DependsOn` property here so I suggest removing it altogether to avoid confusion.
The only dependency is on the `schema` and that's already ensured by the `Schema` arg.

antoninbas (commented):
You think that defining a custom format is neater than just using `FIELD_OPTIONALLY_ENCLOSED_BY` in the `COPY` statement? I tend to agree as it is easier to manage if we need extra customizations in the future.

snowflake/pkg/infra/stack.go
dreamtalen: I didn't add any dependencies at first, but encountered an error when creating the pipe said `File format 'FLOWFORMAT' does not exist or not authorized.`, althrough previous log has shown `FileFormat antrea-sf-flows-format created`. After testing serveral rounds, I found adding these two dependencies could solve this issue.
I agree this looks confusing, do you have any suggestions I could try?

antoninbas: It's a bit of a shot in the dark but first, I'd recommend using the full file format name ("<db>.<schema>.<name>") in the copy statement. I think that by adding these dependencies, you may be affecting session parameters and hiding a bug. Just like I am using the full name for the stage I think you need the full stage for the file format.

dreamtalen: Thanks for advices! I just tested out the simpler way (adding that option in copy into statement directly). Think twice I agree it would be better to have a custom file format if we have multiple customizations to do in the future. For now it may be a little over engineering.

antoninbas: > I agree it would be better to have a custom file format if we have multiple customizations to do in the future. For now it may be a little over engineering.

The thing is we have to keep the upgrade process robust. Pulumi is supposed to help with that, but the Snowflake provider is not always perfect. Switching from inline options to separate file format may be a bit tricky. I'll let you decide as I think additional customizations are not that likely. I still think you should give a try to the full name solution, it may just fix the issue.

dreamtalen: Yes the full name solution solves that issue. Sorry I only edited my last comment and didn't send a new one.

antoninbas (approved):
LGTM, assuming you no longer want to introduce a separate file format resource

Issue #134, opened 2022-10-27
Store recommended policies individually


salv-orlando:
Up to Theia 0.3  the full output of a Spark Policy Recommendation job is stored in a single record.
We should have distinct records for each recommended policy, and handle corresponding data migrations.
The Theia manager controller for recommended policies should also be amended to reflect this change.

PR #138, opened 2022-11-01
Support structured policy recommendation results

yanjunz97:
This PR adds support for structured policy recommendations with the following changes:

- Updates ClickHouse data schema to store each PR individually
- Removes RecommendedNetworkPolicy CRD
- Adds PR results retrieving in the API server
- Adds e2e tests to verify the table recommendations during upgrading

Fix: #134 


pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
```diff
@@ -60,6 +69,15 @@ func (r *REST) Get(ctx context.Context, name string, options *metav1.GetOptions)
 	}
 	intelliNPR := new(intelligence.NetworkPolicyRecommendation)
 	r.copyNetworkPolicyRecommendation(intelliNPR, npReco)
+	// Try to retrieve result from ClickHouse in case NPR is completed
+	if npReco.Status.State == crdv1alpha1.NPRecommendationStateCompleted {
```
yuntanghsu, line 73: Do you think we need to add this to List() as well?

yanjunz97: It seems we only display `CreationTime, CompletionTime, Name, Status` for the list command in CLI. Maybe we can skip result retrieving in the list API to save time?

yuntanghsu: Since get() and list() are similar, will it be better to have consistency for these two functions? I don't have strong opinion. As you said, we only display these 4 columns when using list command.

yanjunz97: Got it. Yes, it makes more sense to keep the consistency considering the generality of the API. I updated the list function. Thanks!

yuntanghsu (approved):
LGTM

build/yamls/flow-visibility.yml
```diff
@@ -152,6 +152,40 @@ data:
     ADD COLUMN clusterUUID String;
     ALTER TABLE flows_local
     ADD COLUMN clusterUUID String;
+  000003_0-3-0.down.sql: |
```
salv-orlando, line 155: Is this a copy/paste of the migrations scripts, or do we have some generation code that automatically populates flow-visibility.yml?

(I think I might have already asked this question - in that case please accept my apologies)

pkg/apis/intelligence/v1alpha1/types.go
```diff
@@ -41,14 +41,14 @@ type NetworkPolicyRecommendation struct {
...
-	TotalStages              int         `json:"totalStages,omitempty"`
-	RecommendedNetworkPolicy string      `json:"recommendedNetworkPolicy,omitempty"`
-	ErrorMsg                 string      `json:"errorMsg,omitempty"`
-	StartTime                metav1.Time `json:"startTime,omitempty"`
-	EndTime                  metav1.Time `json:"endTime,omitempty"`
+	State                 string      `json:"state,omitempty"`
+	SparkApplication      string      `json:"sparkApplication,omitempty"`
+	CompletedStages       int         `json:"completedStages,omitempty"`
+	TotalStages           int         `json:"totalStages,omitempty"`
+	RecommendationOutcome string      `json:"recommendationOutcome,omitempty"`
```
salv-orlando, line 48: forgive me for this dumb question, but should RecommendationOutcome be also int he CRD descriptor?

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
```diff
@@ -60,6 +69,15 @@ func (r *REST) Get(ctx context.Context, name string, options *metav1.GetOptions)
...
+		result, err := r.getRecommendationResult(npReco.Status.SparkApplication)
+		if err != nil {
+			klog.Warningf("Failed to get the result for completed NetworkPolicy Recommedation with id %s, error: %v", npReco.Status.SparkApplication, err)
```
salv-orlando, line 76: do you think it might be worth populating ErrorMsg in the CRD status if we fail to fetch the results from CH?

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
```diff
@@ -153,11 +180,34 @@ func (r *REST) copyNetworkPolicyRecommendation(intelli *intelligence.NetworkPoli
...
-	if crd.Status.RecommendedNP != nil {
-		intelli.Status.RecommendedNetworkPolicy = crd.Status.RecommendedNP.Spec.Yamls
-	}
 	intelli.Status.ErrorMsg = crd.Status.ErrorMsg
 	intelli.Status.StartTime = crd.Status.StartTime
 	intelli.Status.EndTime = crd.Status.EndTime
 	return nil
 }
+
+func (r *REST) getRecommendationResult(id string) (result string, err error) {
```
salv-orlando, line 189: This is more a design question, but do you think there is a need to query clickhouse for every GET? Or once the results are populated in the resource status we can skip the CH query?

The results are kind of static, unless I'm missing something obvious we will get the same data for every query...

salv-orlando (commented):
Partial review. Planning to complete it today.

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -372,50 +338,15 @@ func (c *NPRecommendationController) updateResult(npReco *crdv1alpha1.NetworkPol
...
-			}
-			klog.V(2).InfoS("Created RecommendedNetworkPolicy", "RecommendedNetworkPolicy", recommendedNetworkPolicy.Name, "NetworkRecommendationPolicy", npReco.Name)
-		}
-		return c.updateNPRecommendationStatus(npReco, crdv1alpha1.NetworkPolicyRecommendationStatus{
-			State:         crdv1alpha1.NPRecommendationStateCompleted,
-			RecommendedNP: recommendedNetworkPolicy,
-			EndTime:       metav1.NewTime(time.Now()),
-		})
+	// Delete related SparkApplication CR
+	err := deleteSparkApplicationIfExists(c.kubeClient, npReco.Namespace, npReco.Status.SparkApplication)
```
salv-orlando, line 342: This is not related to your PR, so please take this comment as optional or a hint for a different PR.
It's not very efficient that we list all spark applications before finding the one to delete. In a large scale env we can have thousands of them. If deletion fails because the app does not exists, can we just handle the failure (and ignore it)?

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -372,50 +338,15 @@ func (c *NPRecommendationController) updateResult(npReco *crdv1alpha1.NetworkPol
...
+	if err != nil {
+		return fmt.Errorf("fail to delete Spark Application: %s, error: %v", npReco.Status.SparkApplication, err)
 	}
-	return nil
+	return c.updateNPRecommendationStatus(npReco, crdv1alpha1.NetworkPolicyRecommendationStatus{
+		State:   crdv1alpha1.NPRecommendationStateCompleted,
+		EndTime: metav1.NewTime(time.Now()),
```
salv-orlando, line 348: minor comment (also rather pedant) Should EndTime be the time.Now(), or the time at which the Spark Application completed?

plugins/policy-recommendation/policy_recommendation_job.py
```diff
@@ -811,13 +824,22 @@ def write_recommendation_result(
...
-        "type": recommendation_type,
-        "timeCreated": datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
-        "yamls": "---\n".join(filter(None, result)),
-    }
-    result_df = spark.createDataFrame([result_dict])
+
+    result_dict_list = []
+    for key, value in result.items():
+        for item in value:
+            if item is not None:
```
salv-orlando, line 831: minor python nit: unless you explicitly need to compare with None, you can just use "if item:"

salv-orlando (commented):
Just some additional, minor comments.

build/yamls/flow-visibility.yml
yanjunz97: `flow-visibility.yml` is generated by running `make manifest`

pkg/apis/intelligence/v1alpha1/types.go
yanjunz97: I think this is the intelligence APIs instead of a CRD. We need this because it defines the contents in the response to CLI.

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
yanjunz97: Yes, I think this is a tradeoff between space and time. As we adopt the structured data in ClickHouse and remove RecommendedNetworkPolicy CRD, I think we do not plan to save the data both in ClickHouse and etcd? I also hesitate with saving the same data from database in the API server. Do you prefer to have a simple cache in the API server to avoid the repeated GET?

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
yanjunz97: I updated to populate the ErrorMsg. My only concern is usually we keep consistency between `status.State` and `status.ErrorMsg`, i.e. `status.ErrorMsg` will be empty if the `status.State` is not failed. But in the case of failing to fetch the results from CH, we need to choose between

- break the consistency, i.e. we may have a State of Completed with some ErrorMsg
- overwrite the State value which indicates the status of the SparkApplication 

Currently I choose the first option. Please tell me if you have other idea for this.

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: That makes sense from the efficiency perspective. Updated, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: I think this depends on what the EndTime refers to. I use `time.Now()` as I think it is the EndTime of the NPR CR, as once the State is changed to Completed we will not update this CR any more. But I do not have a strong preference if SparkApplication complete time is more valuable to be set as the EndTime.

plugins/policy-recommendation/policy_recommendation_job.py
yanjunz97: Updated, thanks!

pkg/apis/intelligence/v1alpha1/types.go
salv-orlando: My question was about whether we needed this attribute must be explicitly specified in the CRD as well. We clarified this is not the case, thanks!

pkg/apiserver/registry/intelligence/networkpolicyrecommendation/rest.go
salv-orlando: I believe you are correct - and your explanation makes perfect sense

salv-orlando (approved):
LGTM!

PR #141, opened 2022-11-28
Fix timeout problem in upgrading test

yanjunz97:
Previously we use the defaultTimeout at 90 second for ClickHouse in Upgrading test.
As ClickHouse Operator takes around 1 minute to restart the ClickHouse Pod, it sometimes
leads to a timeout in the upgrading test. This PR increases the timeout to avoid this problem.


wsquan171 (approved):
LGTM. Thanks for taking care of this issue. 
nit: It seems the old comments On L1226 is redundant now.

dreamtalen (approved):

yanjunz97:
> LGTM. Thanks for taking care of this issue. nit: It seems the old comments On L1226 is redundant now.

Good catch. Remove it!

PR #142, opened 2022-11-29
Grafana Service Dependency Graph

Dhruv-J:
Grafana Service Dependency Graph Plugin allows users to visualize how pods send and receive traffic to and from other pods and services.

Signed-off-by: Dhruv-J <dhruvj@vmware.com>


Dhruv-J:
Grafana Service Dependency Graph Plugin allows users to visualize how pods send
and receive traffic to and from other pods and services. To help visualize the
flows, the plugin groups pods by pod label and shows them 'inside' the node they
are a part of. A line pointing from source to destination is also shown, with
the amount of data transmitted labelling the line.

Service Dependency Graph Dashboard

Users can filter graph results based on the number of flow records, as well as
the source and destination pod names.

Service Dependency Graph Dashboard

build/charts/theia/values.yaml
```diff
@@ -151,6 +151,7 @@ grafana:
   installPlugins:  
     - https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.2.zip;theia-grafana-sankey-plugin
     - https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-chord-plugin-1.0.1.zip;theia-grafana-chord-plugin
+    - https://file.io/HWfwKx57UxHq;theia-grafana-dependency-plugin
```
heanlan, line 154: I tried to deploy and test it by myself and got the following error in Grafana Pod log:
```
Error: ✗ failed to download plugin archive: {"success":false,"status":404,"code":"FILE_NOT_FOUND","message":"File not found","help":"https://file.io/help/api/errors/FILE_NOT_FOUND","key":null} (Grafana v8.3.3 linux-amd64)
```

The plugin link gives me: 
image

Maybe you can consider make a fake github release, upload the plugin as a github release assets and use the link here. e.g. https://github.com/heanlan/grafana-chord-plugin/releases/download/1.0.1/grafana-chord-plugin-1.0.0.zip

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
```diff
@@ -0,0 +1,146 @@
+# Grafana Service Dependency Graph Plugin
+
+## What is Grafana Service Dependency Graph Plugin?
+
+Grafana Service Dependency Graph Plugin allows users to visualize how pods send
+and receive traffic to and from other pods and services. To help visualize the
+flows, the plugin groups pods by pod label and shows them 'inside' the node they
```
heanlan, line 7:
```suggestion
flows, the plugin groups pods by Pod labels and shows them 'inside' the node they
```

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
```diff
@@ -0,0 +1,146 @@
...
+Clickhouse query example:
+```sql
+SELECT sourcePodName,
+sourcePodLabels,
+sourceNodeName,
+destinationPodName,
+destinationPodLabels,
+destinationNodeName,
+destinationServicePortName,
+octetDeltaCount
```
heanlan, line 60: Two questions on the query:

1. shall we add a time filter on the query? please refer to other dashboards [queries](https://github.com/antrea-io/theia/blob/b6c1c1f181853a7a4edc9f60769970c2fb0729e3/build/charts/theia/provisioning/dashboards/homepage.json#L119) to see how we add it
2. shall we do `sum(octetDeltaCount)` among the selected time range? otherwise it doesn't make much sense to me by showing the standalone octetDeltaCount value

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
```diff
@@ -0,0 +1,146 @@
...
+FROM flows
+WHERE ( destinationPodName IS NOT NULL AND destinationPodName != '' AND destinationPodName != 'undefined' )
+AND ( sourcePodName IS NOT NULL AND sourcePodName != '' AND sourcePodName != 'undefined' )
+AND ( positionCaseInsensitiveUTF8('${querySourcePodNamespace:raw}', sourcePodNamespace) > 0 )
```
heanlan, line 64: what's the reason for not simply do `sourcePodNamespace in (${querySourcePodNamespace:raw})`

build/charts/theia/provisioning/dashboards/service_dependency_graph_dashboard.json
```diff
@@ -0,0 +1,220 @@
...
+      }
+    ]
+  },
+  "time": {
+    "from": "now-6h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "service_dependency_graph",
```
heanlan, line 216: this should be in `xxxx_dashboard` format

heanlan (commented):
General comments:

- please update the plugin link to a valid one so we can test it
- please remove unnecessary files in the plugin folder, e.g. `plugins/grafana-custom-plugins/grafana-dependency-plugin/package-lock.json`. You can refer to other plugins folder to check which files we decided to keep
- I feel we can generate a better name compare to `service_dependency_graph_dashboard`. I'm curious on the name, e.g. why "service"? what is the "service" referring to? why "dependency"? I searched "service dependency graph" online, didn't see it's a commonly-used name to describe network topology. Could you share more details about why you choose this name. Otherwise, I would prefer something like "network_topology_dashboard"
- Meanwhile, I think "theia-grafana-dependency-plugin" is acceptable as a plugin name. As the plugin can be used on other kinds of data other than network flows.

build/charts/theia/provisioning/dashboards/service_dependency_graph_dashboard.json
```diff
@@ -0,0 +1,220 @@
...
+                  "${queryPodName:raw}"
+                ]
+              }
+            ],
+            "mode": "list",
+            "table": "flows"
+          },
+          "datasource": {
+            "type": "grafana-clickhouse-datasource",
+            "uid": "9JmyLZOVk"
```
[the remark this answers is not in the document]
Dhruv-J: In testing it was the same, but it may have changed after I deleted the grafana-clickhouse-datasource folder. I'll manually change it back to the original.

docs/network-flow-visibility.md
```diff
@@ -750,6 +751,21 @@ Mouse out or click on the background will bring all the traffic back.
 
 <img src="https://downloads.antrea.io/static/08032022/flow-visibility-np-3.png" width="900" alt="Network-Policy Flows Dashboard">
 
+#### Service Dependency Graph Dashboard
```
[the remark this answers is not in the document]
Dhruv-J: That makes sense, will do.

build/charts/theia/values.yaml
Dhruv-J: Sounds good, will update the link.

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
Dhruv-J: 1. yeah, I'll add that as well
2. The plugin reads all individual flows from the query, and they are processed afterwards. The octectDeltaCount itself is never shown, only used to calculate the total which is. If sum() was used, it would give a total of all bytes transferred across all flows, which may not be a helpful metric to the viewer.

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
Dhruv-J: I tried that earlier, it returns one string with comma separated values, which 'in' is unable to use.

build/charts/theia/values.yaml
yanjunz97, line 154: You will always need to run `make` under `build/charts` to update the README file for Helm chart when you update the `value.yaml` file.

yanjunz97 (commented):
Just for curiosity, do you have any idea why we have two `chi-clickhouse-clickhouse-0-0-0`, one in kind-worker and another in kind-worker2 in the first image you share in comment? If it refers to the Pod name, I thought it should show up only once either on kind-worker or kind-worker2?

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
heanlan: thanks for the explanation. got it.

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
heanlan: I think I use it successfully in other dashboards under [snowflake](https://github.com/antrea-io/theia/tree/main/snowflake/grafana/provisioning/dashboards) directory, if you search `clusterUUID` in the dashboard JSON files, you'll be able to see. May be worth looking into what's the difference. It might be related to how you define the filter.

build/charts/theia/values.yaml
heanlan: +1 please also take care of the go test failures https://github.com/antrea-io/theia/actions/runs/3589954800/jobs/6042870930 @Dhruv-J

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
Dhruv-J: Good idea, will look into it.

heanlan:
Hi @Dhruv-J, please also consider add relevant e2e test cases here, to test the newly-added dashboard: https://github.com/antrea-io/theia/blob/b6c1c1f181853a7a4edc9f60769970c2fb0729e3/test/e2e/util.go#L92
there shouldn't be much work, as the test framework is already there, we only need to add test cases into `grafanaTestCases `.

heanlan:
```
=== RUN   TestFlowVisibility/IPv4/Grafana
    flowvisibility_test.go:608: 
        	Error Trace:	flowvisibility_test.go:608
        	            				flowvisibility_test.go:529
        	            				flowvisibility_test.go:495
        	Error:      	"{\"schema\":{\"name\":\"A\",\"refId\":\"A\",\"meta\":{\"executedQueryString\":\"SELECT sourcePodName, sourcePodLabels, sourceNodeName, destinationPodName, destinationPodLabels, destinationNodeName, destinationServicePortName, octetDeltaCount FROM flows\\nWHERE   ( destinationPodName IS NOT NULL AND destinationPodName != '' AND destinationPodName != 'undefined' )\\nAND ( sourcePodName IS NOT NULL AND sourcePodName != '' AND sourcePodName != 'undefined' )\\nAND ( octetDeltaCount IS NOT NULL AND octetDeltaCount != 0)\\nAND ( sourcePodNamespace IN ${querySourcePodNamespace:raw} )\\nAND ( destinationPodNamespace IN ${queryDestinationPodNamespace:raw} )\\nAND ( CAST(flowType AS varchar) IN ${queryFlowType:raw} )\\nAND flowEndSeconds \\u003e= '1670536087' AND flowEndSeconds \\u003c= '1670536987'\\nLIMIT ${queryNumFlows}\"},\"fields\":[]},\"data\":{\"values\":[]}}" does not contain "testing"
```
seems like you successfully provisioned the dashboard to Grafana, just the query result is empty. You may want to keep the test setting not cleanup, login Grafana and check it out.

salv-orlando:
As the review process is still ongoing, we will move this PR to the next milestone.

yanjunz97:
Hi @Dhruv-J , I’ve tried the plugin locally, do you know why I will get the text instead of the graph like the following?
Screen Shot 2022-12-22 at 10 39 16 AM

I also notice sometime I will get something like the following once I open the page.
Screen Shot 2022-12-22 at 10 35 21 AM

yanjunz97:
I still got the same error as before. I just installed antrea, flow-aggregator, and flow-visibility with default settings. @Dhruv-J Do we need to do anything else to make it work?

heanlan:
> I still got the same error as before. I just installed antrea, flow-aggregator, and flow-visibility with default settings. @Dhruv-J Do we need to do anything else to make it work?

+1
image

heanlan:
Hi @Dhruv-J , I have tested your latest commit. I am able to see the chart rendered out. A few questions:

- Have you tried the filters you added at the top? They are not working on my side. I can only see "All", without seeing other values.
image

- When the destinationServicePortName is empty, it gives `svc_` in the chart, which I suppose should be filtered out?
image (3)

- Is there something not working correctly when parsing the Pod names? I have created several Pods named by web-client1-1, web-client1-2, web-client3-1, etc. The names looks correct in the table view, but were shortened to web-client-1, web-client-3 in the chart.
image (1)
image (2)

heanlan:
> Since I have to sanitize the data to get the simpler names, the sourcePodName still shows all unique Pod names in the flow record. 

could you expand a little bit on it? What does it mean by sanitize to get simpler names? @Dhruv-J

Dhruv-J:
> > Since I have to sanitize the data to get the simpler names, the sourcePodName still shows all unique Pod names in the flow record.
> 
> could you expand a little bit on it? What does it mean by sanitize to get simpler names? @Dhruv-J

Yeah, currently, the plugin checks if Pods have label values for the app name, and if not, uses the Pod name as the effective name. While getting the label values, the plugin converts the JSON string of labels to a dictionary and checks for keys 'app' or 'k8s-app'. It then sets the effective name to the corresponding value of one of those keys. I'm not sure if this would be possible in SQL, which is why the sourcePodName variable still shows the normal Pod names.

heanlan:
Could you give me some context why do we only look for labels with key "app" or "k8s-app"? What if the key is "app.kubernetes.io/name" or other self-defined formats?

I believe by default we'd better to display Pods by their Pod names. Like you suggested in the meeting, we can expose an option to users so let them see the flows aggregated by labels. This can be done in a follow-up PR, as I expect it will require some design review.

Dhruv-J:
> Could you give me some context why do we only look for labels with key "app" or "k8s-app"? What if the key is "app.kubernetes.io/name" or other self-defined formats?
> 
> I believe by default we'd better to display Pods by their Pod names. Like you suggested in the meeting, we can expose an option to users so let them see the flows aggregated by labels. This can be done in a follow-up PR, as I expect it will require some design review.

Initially I was checking those two labels, because every Pod on my local cluster had one of the two labels. I didn't have self-defined formats in mind, but it is something I will need to account for. I will set it to just Pod names for now, and keep the label feature in mind for later.

heanlan:
- The filter is broken:

image

It is not well defined in Settings:

image

You can follow exactly what we did for other dashboards.

image

- And I can still observe the empty service issue:

image

build/charts/theia/provisioning/dashboards/network_topology_dashboard.json
```diff
@@ -0,0 +1,156 @@
...
+          "format": 1,
+          "meta": {
+            "builderOptions": {
+              "fields": [],
+              "limit": 100,
+              "mode": "list"
+            }
+          },
+          "queryType": "sql",
+          "rawSql": "SELECT sourcePodName, sourcePodLabels, sourceNodeName, destinationPodName, destinationPodLabels, destinationNodeName, destinationServicePortName, octetDeltaCount FROM flows\nWHERE   ( destinationPodName IS NOT NULL AND destinationPodName != '' )\nAND ( destinationServicePortName IS NOT NULL AND destinationServicePortName != '' )\nAND ( sourcePodName IS NOT NULL AND sourcePodName != '' )\nAND ( octetDeltaCount IS NOT NULL AND octetDeltaCount != 0)\nAND $__timeFilter(flowEndSeconds)\nLIMIT ${numFlows}",
```
heanlan, line 61: If you do not order by `flowEndSeconds`, the `LIMIT ${numFlows}` will not makes much sense?

You might want to sort it by time so you can claim we are displaying the latest xxx flows.

heanlan (commented):
the filter is still not correctly set...

 It should be a constant type named clickhouse_adhoc_query, like:
Pasted Graphic 1

not ad-hoc filter:
Pasted Graphic 2

Otherwise it will populate ALL the tables and columns, not just table flows. Could you see the difference:
image
image

documentation: https://github.com/grafana/clickhouse-datasource#ad-hoc-filters

Dhruv-J:
> the filter is still not correctly set...
> 
> It should be a constant type named clickhouse_adhoc_query

Oh ok, will fix.

test/e2e/util.go
```diff
@@ -316,6 +317,17 @@ var grafanaTestCases = []struct {
...
 		},
 	},
+	{
+		dashboardName: "network_topology_dashboard",
+		dashboardUid:  networkTopologyDashboardUid,
+		queryList: []query{
+			{
+				queryId:      0,
+				panelName:    "Network Topology",
+				expectResult: []string{controlPlaneNodeName(), workerNodeName(1), toExternalClientName, toExternalServerIP},
```
heanlan, line 327: 1. I'm confused. Why do `toExternalClientName, toExternalServerIP` appear in the response string? Doesn't the panel query rule out to-external traffic by requiring "destinationPodName" non-empty?
2. could we add more pod and service names to check? e.g. "perftest-a", "perftest-b", "perftest-c", "perftest-d", "perftest-e", "antrea-test/perftest-b", "antrea-test/perftest-c"

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,31 @@ Mouse out or click on the background will bring all the traffic back.
...
 <img src="https://downloads.antrea.io/static/08032022/flow-visibility-np-3.png" width="900" alt="Network-Policy Flows Dashboard">
 
+#### Network Topology Dashboard
+
+Network Topology Dashboard visualizes Pod-to-Pod and Pod-to-Service traffic,
+including intra-Node traffic as well. Pods labels are visually grouped by Node.
+Each arrow points from the source to the destination, and is labelled by the
+amount of data transmitted.
+
+<img src="" width="900" alt="Network Topology Dashboard">
```
heanlan, line 765: could you fill in the img src?

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,31 @@ Mouse out or click on the background will bring all the traffic back.
 
 <img src="https://downloads.antrea.io/static/08032022/flow-visibility-np-3.png" width="900" alt="Network-Policy Flows Dashboard">
 
+#### Network Topology Dashboard
+
+Network Topology Dashboard visualizes Pod-to-Pod and Pod-to-Service traffic,
+including intra-Node traffic as well. Pods labels are visually grouped by Node.
```
heanlan, line 761: IMO "including intra-Node traffic as well" doesn't form a mutually exclusive relationship with "Pod-to-Pod" and "Pod-to-Service" flows. I suppose we can just remove it.

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,31 @@ Mouse out or click on the background will bring all the traffic back.
...
+Each arrow points from the source to the destination, and is labelled by the
+amount of data transmitted.
```
heanlan, line 763:
```suggestion
cumulative amount of data transmitted in the selected time range.
```

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,31 @@ Mouse out or click on the background will bring all the traffic back.
...
+
+<img src="" width="900" alt="Network Topology Dashboard">
+
+Users can filter graph results based on the number of flow records, as well as
+the source and destination pod names.
+
+We can filter on desired fields, e.g. for number of flow records:
```
heanlan, line 770: redundant line?

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,31 @@ Mouse out or click on the background will bring all the traffic back.
...
+#### Network Topology Dashboard
+
+Network Topology Dashboard visualizes Pod-to-Pod and Pod-to-Service traffic,
+including intra-Node traffic as well. Pods labels are visually grouped by Node.
+Each arrow points from the source to the destination, and is labelled by the
+amount of data transmitted.
+
+<img src="" width="900" alt="Network Topology Dashboard">
+
+Users can filter graph results based on the number of flow records, as well as
```
heanlan, line 767:
```suggestion
Users can filter graph results based on the number of flow records.
```
not just source and destination Pod names but all the fields? we can just remove it, as the same filter has been introduced in the above dashboard.

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,31 @@ Mouse out or click on the background will bring all the traffic back.
...
+We can filter on desired fields, e.g. for number of flow records:
+
+- 100
+- 1000
+- 2500
+- 5000
+- 7500
+- 10000
+
+Here is an example if we filter to show the last 1000 flows.
```
heanlan, line 779:
```suggestion
Here is an example if we filter to show the latest 1000 flows.
```

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,31 @@ Mouse out or click on the background will bring all the traffic back.
...
+
+<img src="" width="900" alt="Network Topology Dashboard">
```
heanlan, line 781: missing img src

could you set the alt texts more accurate for each image rather than using the same texts for all images?

heanlan (commented):
we can add UT for this plugin in a follow-up PR

docs/network-flow-visibility.md
Dhruv-J: Screen Shot 2023-02-22 at 6 44 19 PM
Screen Shot 2023-02-22 at 6 44 51 PM
(just uploading here for the URL)

build/charts/theia/provisioning/dashboards/network_topology_dashboard.json
```diff
@@ -0,0 +1,105 @@
...
+          "format": 1,
+          "meta": {
+            "builderOptions": {
+              "fields": [],
+              "limit": 100,
+              "mode": "list"
+            }
+          },
+          "queryType": "sql",
+          "rawSql": "SELECT sourcePodName, sourceNodeName, destinationPodName, destinationNodeName, destinationServicePortName, octetDeltaCount FROM flows\nWHERE   destinationPodName != ''\nAND sourcePodName != ''\nAND octetDeltaCount != 0\nAND $__timeFilter(flowEndSeconds)\nORDER BY flowEndSeconds DESC",
```
heanlan, line 61: to be consistent to other dashboards, could you add the following filter to the sql query:
```sql
WHERE sourcePodNamespace NOT IN ('kube-system', 'flow-visibility', 'flow-aggregator')
    AND destinationPodNamespace NOT IN ('kube-system', 'flow-visibility', 'flow-aggregator')
```
this is to hide all the system-level default flows. So that users can focus on workload flows in the cluster.

I just feel the default flows are already a bit overwhelming↓ better to hide them. But I don't have a strong opinion. What do you think?
image

heanlan, line 61: remove extra space?
```suggestion
          "rawSql": "SELECT sourcePodName, sourceNodeName, destinationPodName, destinationNodeName, destinationServicePortName, octetDeltaCount FROM flows\nWHERE destinationPodName != ''\nAND sourcePodName != ''\nAND octetDeltaCount != 0\nAND $__timeFilter(flowEndSeconds)\nORDER BY flowEndSeconds DESC",
```

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
```diff
@@ -0,0 +1,144 @@
...
+to return the following fields, in arbitrary order.
+
+- field 1: sourcePodName value with name or an alias of `sourcePodName`
+- field 2: sourceNodeName value with name or an alias of `sourceNodeName`
+- field 3: destinationPodName value with name or an alias of `destinationPodName`
+- field 4: destinationNodeName value with name or an alias of `destinationNodeName`
+- field 5: destinationServicePortName value with name or an alias of `destinationServicePortName`
+- field 6: octetDeltaCount value with name or an alias of `octetDeltaCount`
+
+Clickhouse query example:
```
heanlan, line 51:
```suggestion
ClickHouse query example:
```

test/e2e/util.go
```diff
@@ -316,6 +317,17 @@ var grafanaTestCases = []struct {
...
 		},
 	},
+	{
+		dashboardName: "network_topology_dashboard",
+		dashboardUid:  networkTopologyDashboardUid,
+		queryList: []query{
+			{
+				queryId:      0,
+				panelName:    "Network Topology",
+				expectResult: []string{controlPlaneNodeName(), workerNodeName(1), "perftest-a", "perftest-b", "perftest-c", "perftest-d", "perftest-e"},
```
heanlan, line 327: could you also add service names? as mentioned in: https://github.com/antrea-io/theia/pull/142#discussion_r1115077975

heanlan:
In the future, when resolving comments in each round of reviews, instead of doing `commit --amend` and `push --force`, perhaps better to submit them in separate commits. Otherwise, reviewers are not able to track what have been changed since last commit.

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,15 @@ Mouse out or click on the background will bring all the traffic back.
...
 <img src="https://downloads.antrea.io/static/08032022/flow-visibility-np-3.png" width="900" alt="Network-Policy Flows Dashboard">
 
+#### Network Topology Dashboard
+
+Network Topology Dashboard visualizes Pod-to-Pod and Pod-to-Service traffic.
+Pods labels are visually grouped by Node. Each arrow points from the source to
+the destination, and is labelled by the cumulative amount of data transmitted
+in the selected time range.
+
+<img src="https://user-images.githubusercontent.com/10016630/220810363-96de054a-fee4-4e94-8c06-26d204f4c166.png" width="900" alt="Network Topology Dashboard service dependency graph">
```
heanlan, line 765: If you agree to hide the system-level default flows, better to replace this screenshot with some self-generated workload flows.

you can use this one:
```
kubectl label nodes k8s-node-control-plane app=control-plane
kubectl apply -f https://raw.githubusercontent.com/heanlan/antrea/e4c13b4c6788f1f1d75b4f412399eebba5a6c750/build/yamls/traffic3.yml
```

build/charts/theia/provisioning/dashboards/network_topology_dashboard.json
Dhruv-J: Yeah, that sounds good. I can also add a toggle for them in the same PR that fixes Issue #164.

Dhruv-J:
> In the future, when resolving comments in each round of reviews, instead of doing `commit --amend` and `push --force`, perhaps better to submit them in separate commits. Otherwise, reviewers are not able to track what have been changed since last commit.

Ah ok, I'll leave the squash for after the LGTM comments then.

build/charts/theia/provisioning/dashboards/network_topology_dashboard.json
```diff
@@ -0,0 +1,105 @@
...
+    "from": "now-30m",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "network_topology_dashboard",
+  "uid": "yRVDEad4k",
+  "version": 1,
+  "weekStart": ""
+}
```
yanjunz97, line 105: Add a new line at the end of the file

plugins/grafana-custom-plugins/grafana-dependency-plugin/src/plugin.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    "screenshots": [],
+    "version": "%VERSION%",
+    "updated": "%TODAY%"
+  },
+  "dependencies": {
+    "grafanaDependency": ">=7.0.0",
+    "plugins": []
+  }
+}
```
yanjunz97, line 35: ditto

yanjunz97 (commented):
For curiosity, is that possible to set the default Grafana theme to Light or set a light color background for the whole graph? As I think the black line outside the yellow node box is a little bit hard to distinguish.
Screen Shot 2023-03-01 at 12 39 07 PM

Dhruv-J:
> For curiosity, is that possible to set the default Grafana theme to Light or set a light color background for the whole graph? 

I think it should be already, but I'll double check.

build/charts/theia/provisioning/dashboards/network_topology_dashboard.json
Dhruv-J: ack

yanjunz97 (approved):
LGTM

docs/network-flow-visibility.md
```diff
@@ -754,6 +755,15 @@ Mouse out or click on the background will bring all the traffic back.
...
 <img src="https://downloads.antrea.io/static/08032022/flow-visibility-np-3.png" width="900" alt="Network-Policy Flows Dashboard">
 
+#### Network Topology Dashboard
+
+Network Topology Dashboard visualizes Pod-to-Pod and Pod-to-Service traffic.
+Pods labels are visually grouped by Node. Each arrow points from the source to
+the destination, and is labelled by the cumulative amount of data transmitted
+in the selected time range.
+
+<img src="https://downloads.antrea.io/static/03012023/flow-visibility-network-topology-0.png" width="900" alt="Network Topology Dashboard service dependency graph">
```
heanlan, line 765: could you adjust the width to a smaller value? Currently it's too big, I need to scroll down to see the entire picture.

heanlan (commented):
one quick comment:
could you also add a line for the new dashboard in the below panel in home dashboard?
image

build/charts/theia/provisioning/dashboards/homepage.json
```diff
@@ -576,7 +576,7 @@
...
...low Records Dashboard</span> displays the flow records being captured in the selected\ntime range.\n\n<span style=\"color:rgb(184, 119, 217)\">Pod-to-Pod Flows Dashboard</span> shows cumulative bytes and throughput of Pod-to-Pod\ntraffic.\n\n<span style=\"color:rgb(184, 119, 217)\">Pod-to-External Flows Dashboard</span> shows cumulative bytes and throughput of\ntraffic from in-cluster to out-of-cluster. The destination of such a traffic flow\nis represented by the destination IP address.\n\n<span style=\"color:rgb(184, 119, 217)\">Pod-to-Service Flows Dashboard</span> shows cumulative bytes and throughput of\nPod-to-Service flows. The destination of such a traffic is represented by the\ndestination Service metadata.\n\n<span style=\"color:rgb(184, 119, 217)\">Node-to-Node Flows Dashboard</span> visualizes the Node-to-Node traffic, including intra-Node\nand inter-Node flows.\n\n<span style=\"color:rgb(184, 119, 217)\">Network-Policy Flows Dashboard</span> visualizes both the traffic with NetworkPolicies enforced,\nand unprotected traffic.\n\n<span style=\"color:rgb(184, 119, 217)\">Network Topology Dashboard</span> visualizes both Pod-to-Pod and Pod-to-Service traffic via a graph.",
```
heanlan, line 579: "The following seven ...

heanlan (approved):

PR #147, opened 2023-01-06
Update the getting-started doc to include Theia Manager

yanjunz97:
The commit update doc to include Theia Manager, as the CLI will be broken without Theia Manager enabled.
It also sets Theia Manager as enabled by default.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -72,20 +72,33 @@ helm repo add antrea https://charts.antrea.io
 helm repo update
 ```
 
-To enable both [Grafana Flow Collector](network-flow-visibility.md) and
-[NetworkPolicy Recommendation](networkpolicy-recommendation.md), please install
-Flow Aggregator and Theia by runnning the following commands:
+To install Flow Aggregator, please run the following command:
 
 ```bash
 helm install flow-aggregator antrea/flow-aggregator --set clickHouse.enable=true,recordContents.podLabels=true -n flow-aggregator --create-namespace
+```
+
+To enable both [Grafana Flow Collector](network-flow-visibility.md) and
+[NetworkPolicy Recommendation](networkpolicy-recommendation.md), please install
+Theia by runnning the following commands:
+
+```bash
 helm install theia antrea/theia --set sparkOperator.enable=true -n flow-visibility --create-namespace
 ```
 
-To enable only Grafana Flow Collector, please install Flow Aggregator and Theia
-by runnning the following commands:
+From Theia v0.3, [Theia Command-line Tool](theia-cli.md) uses Theia Manager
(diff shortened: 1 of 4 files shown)


docs/getting-started.md
```diff
@@ -89,6 +89,15 @@ helm install flow-aggregator antrea/flow-aggregator --set clickHouse.enable=true
 helm install theia antrea/theia -n flow-visibility --create-namespace
 ```
 
+From Theia v0.3, [Theia Command-line Tool](theia-cli.md) uses Theia Manager
+as a layer to connect and manage other resources like ClickHouse and Spark
+Operator. To enable Theia Manager, please run the following command when
+installing Theia:
+
+```bash
+helm install theia antrea/theia --set sparkOperator.enable=true,theiaManager.enable=true -n flow-visibility --create-namespace
```
dreamtalen, line 98: Hi Yanjun, I take a look together with the above paragraphs and found we are missing `helm install flow-aggregator` here. 
Since I feel writing `helm install flow-aggregator antrea/flow-aggregator --set clickHouse.enable=true,recordContents.podLabels=true -n flow-aggregator --create-namespace` three times is a bit repetitive. Could we split the installation section of flow-aggregator and theia in these paragraphs? It may look better IMO.

yanjunz97: I also feel the commands repetitive though I remember the original purpose to put them together is to keep the installation in "one step". Anyway it still sounds good to me to split the step under the current condition. Updated, thanks!

dreamtalen (approved):
LGTM

docs/getting-started.md
```diff
@@ -72,20 +72,33 @@ helm repo add antrea https://charts.antrea.io
...
 ```bash
-helm install flow-aggregator antrea/flow-aggregator --set clickHouse.enable=true,recordContents.podLabels=true -n flow-aggregator --create-namespace
 helm install theia antrea/theia --set sparkOperator.enable=true -n flow-visibility --create-namespace
 ```
 
-To enable only Grafana Flow Collector, please install Flow Aggregator and Theia
-by runnning the following commands:
+From Theia v0.3, [Theia Command-line Tool](theia-cli.md) uses Theia Manager
+as a layer to connect and manage other resources like ClickHouse and Spark
+Operator. To enable Theia Manager, please install Theia by runnning the
```
yuntanghsu, line 91: Do we need to set theiaManager.enable=true as we already set it to true in values.yaml?

yanjunz97: Actually the answer is no. It is not necessary when we set the default to true, while it is no harm to add this. I just have not come up with a better description to indicate only for v0.3 and v0.4 we need to add `theiaManager.enable=true`.

yuntanghsu (approved):
LGTM

PR #150, opened 2023-01-14
Resync Theia Manager at startup

### LINKED ISSUE #135: Resync Theia Manager at startup
If the Theia manager is restarted due to events such as errors, evacuation or user actions we must make sure it synchronizes with the latest state with both Clickhouse and Spark.

From a scale perspective, we should ensure this synchronization does not add significant time to startup processes when there is a large data set to validate.

yanjunz97:
Theia Manager needs to be resynchronized with the SparkApplications and
db entries when it recovers from the downtime. This commit

- Supports removing stale SparkApplications and db entries after Theia Manager restart
- Supports adding back scheduled/running NPR to periodical sync list
- Adds e2e tests for the Theia Manager restart

Fixes #135 

### DIFF
--- a/pkg/controller/networkpolicyrecommendation/controller.go
+++ b/pkg/controller/networkpolicyrecommendation/controller.go
@@ -41,6 +41,7 @@ import (
 	"antrea.io/theia/pkg/client/listers/crd/v1alpha1"
 	"antrea.io/theia/pkg/util"
 	"antrea.io/theia/pkg/util/clickhouse"
+	"antrea.io/theia/pkg/util/env"
 	sparkv1 "antrea.io/theia/third_party/sparkoperator/v1beta2"
 )
 
@@ -69,12 +70,20 @@ var (
 	// Spark Application CRUD functions, for unit tests
 	CreateSparkApplication   = createSparkApplication
 	DeleteSparkApplication   = deleteSparkApplication
-	ListSparkApplication     = listSparkApplication
+	ListSparkApplication     = listSparkApplicationWithLabel
 	GetSparkApplication      = getSparkApplication
 	GetSparkMonitoringSvcDNS = getSparkMonitoringSvcDNS
 	// For NPR in scheduled or running state, check its status periodically
 	npRecommendationResyncPeriod = 10 * time.Second
+	sparkAppLabelMap             = map[string]string{"app": "theia-npr"}
+	sparkAppLabel                = "app=theia-npr"
 )
+
+type gcKey struct {
+	removeStaleDbEntries bool
+	removeStaleSparkApp  bool
+	addResync            bool
+}
 
 type NPRecommendationController struct {
 	crdClient  versioned.Interface
@@ -86,6 +95,7 @@ type NPRecommendationController struct {
 	// queue maintains the Service objects that need to be synced.
 	queue                  workqueue.RateLimitingInterface
 	deletionQueue          workqueue.RateLimitingInterface
+	gcQueue                workqueue.RateLimitingInterface
 	periodicResyncSetMutex sync.Mutex
 	periodicResyncSet      map[apimachinerytypes.NamespacedName]struct{}
 	clickhouseConnect      *sql.DB
@@ -106,6 +116,7 @@ func NewNPRecommendationController(
 		kubeClient:               kubeClient,
 		queue:                    workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(minRetryDelay, maxRetryDelay), "npRecommendation"),
 		deletionQueue:            workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(minRetryDelay, maxRetryDelay), "npRecommendationCleanup"),
+		gcQueue:                  workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(minRetryDelay, maxRetryDelay), "npRecommendationGarbageCollection"),
 		npRecommendationInformer: npRecommendationInformer.Informer(),
 		npRecommendationLister:   npRecommendationInformer.Lister(),
 		npRecommendationSynced:   npRecommendationInformer.Informer().HasSynced,
@@ -186,6 +197,8 @@ func (c *NPRecommendationController) deleteNPRecommendation(old interface{}) {
 // workqueue.
 func (c *NPRecommendationController) Run(stopCh <-chan struct{}) {
 	defer c.queue.ShutDown()
+	defer c.deletionQueue.ShutDown()
+	defer c.gcQueue.ShutDown()
 
 	klog.InfoS("Starting controller", "name", controllerName)
 	defer klog.InfoS("Shutting down controller", "name", controllerName)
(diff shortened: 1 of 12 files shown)


pkg/util/clickhouse/clickhouse.go
```diff
@@ -32,19 +32,19 @@ const (
...
 )
 
 var (
 	openSql         = sql.Open
 	createK8sClient = k8s.CreateK8sClient
 )
 
-func SetupConnection() (connect *sql.DB, err error) {
-	url, err := getClickHouseURL()
+func SetupConnection(client *kubernetes.Interface) (connect *sql.DB, err error) {
```
wsquan171, line 46:
```suggestion
func SetupConnection(client kubernetes.Interface) (connect *sql.DB, err error) {
```
Go interfaces are already pointers. Should avoid using pointers to interfaces

pkg/util/clickhouse/clickhouse.go
```diff
@@ -91,17 +91,22 @@ func GetSecret(client kubernetes.Interface, namespace string) (username string,
 	return username, password, nil
 }
 
-func getClickHouseURL() (url string, err error) {
+func getClickHouseURL(clientPtr *kubernetes.Interface) (url string, err error) {
```
wsquan171, line 94: same here

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -194,6 +199,10 @@ func (c *NPRecommendationController) Run(stopCh <-chan struct{}) {
 		return
 	}
 
+	// The key can be anything as we only have single item.
+	c.gcQueue.Add("key")
```
wsquan171, line 203: Maybe create a const for the key in this package and use that?

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -206,6 +215,95 @@ func (c *NPRecommendationController) Run(stopCh <-chan struct{}) {
...
+				errorList = append(errorList, err)
+			}
+		}
+	}
+	if len(errorList) > 0 {
+		return fmt.Errorf("failed to remove stale Spark Applications and database entries: %v", errorList)
+	}
+
+	// Add scheduled/running NPR back to resycn list
+	nprList, err := c.ListNetworkPolicyRecommendation(env.GetTheiaNamespace())
```
wsquan171, line 269: Could there be a race case when a job is deleted between `ListNetworkPolicyRecommendation` and  `c.addPeriodicSync` is called? Shouldn't be of much issue though as even that's the case, during the next resync the cleanup can be done w/o issue.

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -206,6 +215,95 @@ func (c *NPRecommendationController) Run(stopCh <-chan struct{}) {
 	<-stopCh
 }
 
+// handleStaleResources handles the stale Spark Applications and database entries.
+// It will delete the dangling resources without a matching NetworkPolicyRecommendation
+// and add the running NetworkPolicyRecommendation back to the periodical watch list.
+func (c *NPRecommendationController) handleStaleResources() error {
```
wsquan171, line 221: I kind of think that we should separate sync / retry of the two things we're doing there. Currently populating resync won't be done if errors occur when deleting stale SA or db entries.

test/e2e/policyrecommendation_test.go
```diff
@@ -167,6 +167,54 @@ func TestPolicyRecommendation(t *testing.T) {
...
+	require.NoError(t, err)
+	assert := assert.New(t)
+	assert.Containsf(stdout, "Status of this policy recommendation job is", "stdout: %s", stdout)
+	err = data.podWaitForReady(defaultTimeout, jobName2+"-driver", flowVisibilityNamespace)
+	require.NoError(t, err)
+	_, err = deleteJob(t, data, jobName2)
+	require.NoError(t, err)
+
+	// Check the SparkApplication and database entries of jobName1 do not exist
+	cmd = fmt.Sprintf("kubectl get sparkapplication %s -n flow-visibility", jobName1)
```
wsquan171, line 211: Maybe wraps this and and L215 in some retries just in case theia manager gets slow in removing them?

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: I think it is fine as the PeriodicSync just add the NPR key to the main workqueue, `syncNPRecommendation` will return nil if it does not find the NPR. As the theia manager is on, the deletion process will take care of the SparkApplication, db entires and also PeriodicSyncSet.

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: Currently `handleStaleResources` will be retried as I use a `gcworker` to run it and add back the key upon error. But it's true if deleting stale SA or db entries never succeeds then populating resync will not be executed. Would you prefer to add another worker to take care of the populating resync thing?

wsquan171: Or we can persist which part of GC has successfully finished in `gcworker`, and only retry the part that has not. Also, if everything is done, should `gcworker` quit rather than blocking on the queue (returning false on L299)?

pkg/util/clickhouse/clickhouse.go
yanjunz97: Thanks Shawn for the reviewing! Updated.

pkg/util/clickhouse/clickhouse.go
yanjunz97: Updated, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: Updated the key to a specific gcKey to allow separation, thanks!

pkg/controller/networkpolicyrecommendation/controller.go
yanjunz97: Sounds great, updated!

test/e2e/policyrecommendation_test.go
yanjunz97: Updated, thanks! Also add some retries for npr deletion as I found some failures.

wsquan171 (approved):
LGTM.

pkg/controller/networkpolicyrecommendation/controller.go
```diff
@@ -194,9 +205,14 @@ func (c *NPRecommendationController) Run(stopCh <-chan struct{}) {
...
 
-	go func() {
-		wait.Until(c.resyncNPRecommendation, npRecommendationResyncPeriod, stopCh)
-	}()
+	c.gcQueue.Add(gcKey{
+		removeStaleDbEntries: true,
+		removeStaleSparkApp:  true,
+		addResync:            true,
+	})
+	go wait.Until(c.gcworker, time.Second, stopCh)
```
yuntanghsu, line 213: I think we only add gcKey to gcQueue when theia-manager is started? I'm wondering why do we use wait.Util?
Won't the func processNextGcWorkItem get blocked at line #333?

yanjunz97: As the key will be added back to the queue when there is an error in `handleStaleResources`, I apply the same logic as the other work queues. And as `processNextGcWorkItem` will return false when `handleStaleResources` succeed, the goroutine will finish at that time and won't be blocked. It should be possible to replace this by a simpler while loop, but I'm not sure if it is a better solution.

yanjunz97: Thanks YunTang, you're right. Using wait.Until makes the goroutine running until the stopCh. Updated!

yuntanghsu (approved):
LGTM

PR #170, opened 2023-02-20
Use new Github Action to upload release assets

antoninbas:
Use alexellis/upload-assets instead of actions/upload-release-asset, which is no longer maintained. One advantage is that we can list all assets with a glob expression, so the workflow file is much less verbose.


antoninbas:
This is based on https://github.com/antrea-io/antrea/pull/4638

yanjunz97 (approved):
LGTM

PR #171, opened 2023-02-21
Code Coverage

Dhruv-J:
This PR defines the thresholds at which code coverage is run, and outlines behavior for the comment bot to follow.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,45 @@
+codecov:
+  branch: main
+  require_ci_to_pass: no
+
+comment:
+  layout: "reach,diff,flags,tree"
+  behavior: default
+  require_changes: no
+  require_base: no
+  require_head: no
+  after_n_builds: 1
+  show_carryforward_flags: true
+
+github_checks:
+  annotations: true
+
+coverage:
+  status:
+    patch:
+      default:
+        target: auto
+        threshold: 5%
+        base: auto
+        only_pulls: true
+        flags:
+          - unit-tests
+    project:
+      default:
+        target: auto
+      theia-unit-tests:
+        target: auto
+        flags:
+          - unit-tests
+
+flag_management:
+  default_rules:
+    carryforward: true
+ignore:
+  - "**/testing/*.go"
+  - "**/mock_*.go"
+  - "**/*generate*.go"
+  - "pkg/client"
+  - "**/pkg/client"
+  - "third_party"
+  - "pkg/api"


elton-furtado:
Can you configure all the required flags similar to antrea in codecov.yml and kind.yml

Dhruv-J:
> Can you configure all the required flags similar to antrea in codecov.yml and kind.yml

Yeah, I thought the existing code in this PR should do that, as I've replicated the setup in go.yml for kind.yml, and updated codecov.yml identically for both. However, only the unit-tests flag, which is in go.yml, is working correctly. I'm trying to look into what could be causing it and if there's additional setup I'm missing. It may be worth it to release flag support for just unit-tests initially, and update the flags to include kind-e2e-tests when we reach 60% coverage. This way, any PRs made during the course of the 0.6 release can still be informational and contribute to reaching a higher coverage percentage. The flags will only provide a more visual breakdown, but will not change the actual coverage percentage.

codecov.yml
```diff
@@ -0,0 +1,45 @@
...
+flag_management:
+  default_rules:
+    carryforward: true
+ignore:
+  - "**/testing/*.go"
+  - "**/mock_*.go"
+  - "**/*generate*.go"
+  - "pkg/client"
+  - "**/pkg/client"
+  - "**/*.pb.go"
```
tushartathgur, line 44: Don’t need this check yet

codecov.yml
```diff
@@ -0,0 +1,45 @@
...
+  - "third_party"
```
tushartathgur, line 45: Add line to the end

codecov.yml
```diff
@@ -0,0 +1,45 @@
...
+        flags:
+          - unit-tests
+
+flag_management:
+  default_rules:
+    carryforward: true
+ignore:
+  - "**/testing/*.go"
+  - "**/mock_*.go"
+  - "**/*generate*.go"
```
tushartathgur, line 41: Don’t need to check for pkg/api as well, no tests are written for the folder yet.

tushartathgur (commented):
initial comments for ignoring files

codecov.yml
```diff
@@ -0,0 +1,45 @@
...
+        target: auto
+        threshold: 5%
+        base: auto
+        only_pulls: true
+        flags:
+          - unit-tests
+    project:
+      default:
+        target: auto
+      theia-unit-tests:
```
dreamtalen, line 30: `e2e-tests` and `kind-e2e-tests` are missing here

dreamtalen (approved):
This code coverage yml does not include e2e-tests and kind-e2e-tests flags yet, so please update the commit message, PR description, and PR title to reflect that only unit-tests are supported for now. 
It would also be good to create a separate issue to track the addition of those two missing flags. 
Everything else looks good to me.

Issue #168, opened 2023-02-14
Add the support for the aggregated throughput anomaly detection


dreamtalen:
Expand the scope of throughput anomaly detection beyond individual endpoint-to-endpoint flow, for example:

- Monitor the accumulated inbound/outbound throughput of a specific Pod.
- Monitor the accumulated inbound/outbound throughput of a set of Pods (based on Pod labels).
- Monitor the accumulated inbound/outbound throughput of a specified Service.

Notes:

- Corresponding CLI changes are needed to complete this enhancement.
- Add ignorelist for namespaces as an argument for TAD
- Consistent Throughput and Algocalc datatypes.
- Restructuring Anomalydetector in api to intelligence in api as required by Theia Manager
- Restructuring Rest files of Anomalydetector in apiserver folder.
- Add Theia resync logic for TAD.
- If https://github.com/antrea-io/theia/issues/167 has been finished first, corresponding changes of the real-time detection are needed.

PR #176, opened 2023-03-03
Adding theia manager resync to TAD and restructuring TAD files

tushartathgur:
This PR works does the following:

-   TAD api and rest files are restructured as per the original design of theia-manager
-   Theia Manager resync is added for TAD that helps remove stale DB and spark apps.
-   Consistency of datatypes for thoughput and algocalc for TAD.
-   Add allowlist for namespaces as an argument for TAD

Partially solves #168


test/e2e/util.go
```diff
@@ -468,3 +472,76 @@ func RetrieveJobResult(t *testing.T, data *TestData, cmd string) (stdout string,
...
+	}
+	err = data.podWaitForReady(defaultTimeout, theiaManagerPodName, flowVisibilityNamespace)
+	if err != nil {
+		return fmt.Errorf("error when waiting for Theia Manager %s", theiaManagerPodName)
+	}
+	return nil
+}
+
+func VerifyJobCleaned(t *testing.T, data *TestData, jobName string, tablename string, prefixlen int) error {
+	// Check the SparkApplication and database entries of jobName1 do not exist
```
dreamtalen, line 519: 'jobName1' -> 'jobName'

test/e2e/util.go
```diff
@@ -468,3 +472,76 @@ func RetrieveJobResult(t *testing.T, data *TestData, cmd string) (stdout string,
...
+	err = wait.PollImmediate(defaultInterval, defaultTimeout, func() (bool, error) {
+		cmd := fmt.Sprintf("kubectl get sparkapplication %s -n flow-visibility", jobName)
+		_, stdout, stderr, _ = data.RunCommandOnNode(controlPlaneNodeName(), cmd)
+		if !strings.Contains(stderr, fmt.Sprintf("sparkapplications.sparkoperator.k8s.io \"%s\" not found", jobName)) {
+			// Keep trying
+			return false, nil
+		}
+		cmd = fmt.Sprintf("clickhouse client -q \"SELECT COUNT() FROM %s WHERE id='%s'\"", tablename, jobName[prefixlen:])
+		queryOutput, stderr, err = data.RunCommandFromPod(flowVisibilityNamespace, clickHousePodName, "clickhouse", []string{"bash", "-c", cmd})
+		require.NoErrorf(t, err, "fail to get tadetector from ClickHouse, stderr: %v", stderr)
```
dreamtalen, line 536: Update this error msg with all kinds of spark application instead of 'tadetector' only

test/e2e/util.go
```diff
@@ -468,3 +472,76 @@ func RetrieveJobResult(t *testing.T, data *TestData, cmd string) (stdout string,
...
+	// Simulate the Theia Manager downtime
+	cmd := "kubectl delete deployment theia-manager -n flow-visibility"
+	_, stdout, stderr, err := data.RunCommandOnNode(controlPlaneNodeName(), cmd)
+	if err != nil {
+		return fmt.Errorf("error when running %s from %s: %v\nstdout:%s\nstderr:%s", cmd, controlPlaneNodeName(), err, stdout, stderr)
+	}
+
+	// Delete the first job during the Theia Manager downtime
+	cmd = fmt.Sprintf("kubectl delete %s %s -n flow-visibility", job, jobName1)
+	// Sometimes the deletion will fail with 'error: the server doesn't have a resource type "npr"'
```
dreamtalen, line 486: Update this comment with all kinds of spark application instead of 'npr' only

pkg/controller/anomalydetector/controller.go
```diff
@@ -191,6 +204,134 @@ func (c *AnomalyDetectorController) Run(stopCh <-chan struct{}) {
...
+		return false
+	} else {
+		klog.ErrorS(err, "Error handling stale resources, requeuing it")
+		c.gcQueue.AddRateLimited(updatedKey)
+	}
+	return true
+}
+
+// handleStaleResources handles the stale Spark Applications and database entries.
+func (c *AnomalyDetectorController) handleStaleResources(key controllerutil.GcKey) (updatedKey controllerutil.GcKey, err error) {
```
dreamtalen, line 235: Is it possible to reuse the series of resync functions (handleStaleResources, handleStaleSparkApp, handleStaleSparkApp) for both the anomaly detector and policy recommendation? It would be more efficient if we could avoid duplicating this code every time a new Spark application is involved.
cc. @yanjunz97

tushartathgur: I think we can use interface for when we have to call a common function and not from within the controller, however that might not work here
e.g. if we have two controllers controller1 and controller 2, we can use a same function for both and pass the struct variables and it should work perfectly but we are calling controller function (GetThroughputAnomalyDetector/GetNetworkPolicyRecommendation) within the common function which has a different return type than the other, which I guess won’t be feasible to pass function as an argument to a common function as diff return type would cause issue, so I guess best practice would be to duplicate the code.
Please correct me if I am wrong.

yanjunz97: As the get and list functions for different resources still have the same number of parameter, maybe you can pass `function(xxx) interface{}` and a string (or type) to indicate the resource as arguments and cast the return value according to the resource in the common function.

pkg/controller/util.go
```diff
@@ -231,3 +235,59 @@ func CreateSparkApplication(client kubernetes.Interface, namespace string, spark
 func GetSparkMonitoringSvcDNS(id string, namespace string, sparkPort int) string {
 	return fmt.Sprintf("http://pr-%s-ui-svc.%s.svc:%d", id, namespace, sparkPort)
 }
+
+func HandleStaleDbEntries(clickhouseConnect *sql.DB, client kubernetes.Interface, job, tableName string, ifResourceExists func(string, string) error, id_prefix string) error {
```
dreamtalen, line 239: nit: id_prefix->idPrefix

dreamtalen (approved):
LGTM, a nit only.

yanjunz97 (approved):
LGTM

PR #199, opened 2023-03-28
Resolving security vulnerabilities for Graphana chord plugin

tushartathgur:


plugins/grafana-custom-plugins/grafana-chord-plugin/package.json
```diff
@@ -15,15 +15,19 @@
     "d3-chord": "^3.0.1",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.6",
+    "json5": "2.x",
```
heanlan, line 18: why do we need the specify json5 version?

plugins/grafana-custom-plugins/grafana-chord-plugin/package.json
```diff
@@ -15,15 +15,19 @@
...
+    "loader-utils": "^3.2.1",
     "my-d3": "npm:d3@7.1.0",
     "react": "^16.13.1",
-    "react-dom": "^16.13.1"
+    "react-dom": "^16.13.1",
+    "simple-git": "^3.6.0",
+    "webpack": "^5.72.0"
```
heanlan, line 24: same for webpack

heanlan:
Hi @tushartathgur , thanks for the PR. I just merged 4 PRs made by Dependabot for upgrading `json5`, `webpack`, `loader-utils`, and `simple-git`. Those packages are all indirect-dependencies. I feel we don't have to declare their version requirements in package.json. By default, Dependabot will only edit yarn.lock file, and leave package.json as it was. For simplicity, we can make use of the auto-generated PRs to resolve the alerts.

We can use this PR to resolve the rest of alerts which needs resolving conflicts first, e.g. https://github.com/antrea-io/theia/security/dependabot/100 . What do you think? Apology for giving some bad advice before.

nits: It should be "Grafana" rather than "Graphana" in the PR title

tushartathgur:
> Hi @tushartathgur , thanks for the PR. I just merged 4 PRs made by Dependabot for upgrading `json5`, `webpack`, `loader-utils`, and `simple-git`. Those packages are all indirect-dependencies. I feel we don't have to declare their version requirements in package.json. By default, Dependabot will only edit yarn.lock file, and leave package.json as it was. For simplicity, we can make use of the auto-generated PRs to resolve the alerts.
> 
> We can use this PR to resolve the rest of alerts which needs resolving conflicts first, e.g. [Prototype Pollution in minimist](/antrea-io/theia/security/dependabot/100) . What do you think? Apology for giving some bad advice before.
> 
> nits: It should be "Grafana" rather than "Graphana" in the PR title

I completely agree with you, we should allow auto generated PRs to resolve the alerts. That helps with diminishing the tedious work of unit testing and version calculation.

Issue #164, opened 2023-02-08
Add label toggle to Dependency Plugin


Dhruv-J:
In the current dependency plugin, individual Pod traffic is shown. In instances where multiple replicas of multiple Pods exist, the graph could get cluttered easily and become unreadable. To remedy this, I suggest we add a toggle to the Network Topology Dashboard that allows users to group Pods by Pod labels.

PR #200, opened 2023-03-28
Dependency Plugin Additional Options

Dhruv-J:
This PR allows for users to edit the panel and choose to
group pods by labels or not. The user can also change the
color of the boxes to red, yellow, green, or blue.

Fixes issue #164


build/charts/theia/README.md
```diff
@@ -52,7 +52,7 @@ Kubernetes: `>= 1.16.0-0`
...
 | grafana.image | object | `{"pullPolicy":"IfNotPresent","repository":"projects.registry.vmware.com/antrea/theia-grafana","tag":"8.3.3"}` | Container image used by Grafana. |
-| grafana.installPlugins | list | `["https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.2.zip;theia-grafana-sankey-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-chord-plugin-1.0.1.zip;theia-grafana-chord-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-dependency-plugin-1.0.0.zip;theia-grafana-dependency-plugin","grafana-clickhouse-datasource 1.0.1"]` | Grafana plugins to install. |
+| grafana.installPlugins | list | `["https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.2.zip;theia-grafana-sankey-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-chord-plugin-1.0.1.zip;theia-grafana-chord-plugin","https://github.com/Dhruv-J/grafana-dependency-plugin/archive/refs/tags/pre3.zip;theia-grafana-dependency-plugin","grafana-clickhouse-datasource 1.0.1"]` | Grafana plugins to install. |
```
yuntanghsu, line 55: Just curious, why we change the path here?

Dhruv-J: That's just the link I use for the test release, so if anyone wants to run my PR branch, they can without any additional setup. I change it back before the final merge occurs.

heanlan (commented):
Would you mind adding either more explanation in the PR description or your feature design in the related issue? Including example screenshots of the new feature, for both the pod label toggle and the color choice.

It is not very clear to me
1. how does they look like in Grafana
2. how we expect the pod label toggle to work? why we only look for "k8s"/"k8s-app" labels? 

I thought we had the discussion before on the previous PR: https://github.com/antrea-io/theia/pull/142#issuecomment-1423187242

plugins/grafana-custom-plugins/grafana-dependency-plugin/src/DependencyPanel.tsx
```diff
@@ -44,26 +61,49 @@ export const DependencyPanel: React.FC<Props> = ({ options, data, width, height
...
     const destinationServicePortName = destinationServicePortNames?.values.get(i);
     const octetDeltaCount = octetDeltaCounts?.values.get(i);
 
+    function getName(groupByLabel: boolean, source: boolean, labelJSON: string) {
+      if(!groupByLabel || labelJSON === undefined) {
+        return source ? sourcePodName : destinationPodName;
+      }
+      console.log('attempting to log JSON: '+labelJSON);
+      let labels = JSON.parse(labelJSON);
+      if(labels['app'] !== undefined) {
```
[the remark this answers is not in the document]
Dhruv-J: In order to group by labels, I added support for the two basic ones I saw that had values for most pods. Should I make it check against a user-defined label?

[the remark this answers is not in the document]
heanlan: > the two basic ones I saw that had values for most pods

to clarify, where does the observation come from? which are the "most pods"? like I mentioned in the previous comment, is "app.kubernetes.io/name" less common than those two per your observation?

My point is, I'm not sure about which particular labels should we look for, so I doubt whether it's a good way to design this feature. I would like to take one step back to the motivation and design. For example, I see there are some common labels listed in k8s documentation: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ , but I'm not convinced that we should look for these labels.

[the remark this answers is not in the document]
Dhruv-J: Screen Shot 2023-04-19 at 1 17 59 PM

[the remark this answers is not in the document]
Dhruv-J: Screen Shot 2023-04-19 at 1 23 22 PM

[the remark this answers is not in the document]
heanlan: - Could you also upload the screenshot of the diagram? 
- Does it mean user can only specify one label each time? Could we make use of the ad-hoc filter, showing all the available labels, and letting users to select one of more labels they would like to group by? I'm not sure if it is doable. If you would like to do a design review/have a design doc, we can discuss it in more details.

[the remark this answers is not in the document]
Dhruv-J: I'll update with a screenshot of the diagram as well, but I don't think that an ad-hoc filter would be feasible for two reasons.
1. If pods are grouped by label, their pod label is used per square instead of the pod name itself. If multiple pod labels were used at once, I don't think there would be a good way of grouping them, especially if there are conflicts among labels. For example two pods could both be labelled by the same value for one label, but different values for another.
2. Ad-hoc filter can only check the database columns, anything more specific would require query variables, which cannot be included in the query if the query is to pass end to end testing. Dropdown menu is a great idea, I just don't know how it could be populated without a query variable. I can look into that, but if it doesn't pan out, users may need to manually enter.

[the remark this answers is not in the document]
heanlan: For the first one, I don't have a strong opinion on whether we should enable single label or multiple labels selection. If single label selection would make the graph representation easier, we can go with it.
 > For example two pods could both be labelled by the same value for one label, but different values for another.

Although I didn't get how does the above situation harm.

For the second one, we can remove this query from E2E test, if necessary. I think it worth exploring on query variables, ad-hoc filters, and maybe other Grafana configurations to see if it will make the dropdown menu work. I concern more on whether we can parse the Pod label string and split them into individual key value pairs.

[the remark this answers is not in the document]
Dhruv-J: Yeah, splitting the label is more challenging than getting the rest of the query variable to work, since distinct values of only the key needs to be selected. There isn't a query that can parse the JSON and format output, so it may not be possible to include a dropdown menu. I hope that users wishing to group by label will already have some idea for which labels they wish to group by.
Screen Shot 2023-04-26 at 10 19 38 AM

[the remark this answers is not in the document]
heanlan: why there are two source Pod groups with the same label value "web-client"?

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
```diff
@@ -106,6 +112,16 @@ installed panels will appear. For more information, visit the docs on [Grafana p
 
 ## Customization
 
+### 3. Customize the Panel Options
+
+Users can customize the panel by choosing to group the diagram based a chosen
+label. It is also possible to change the color of the pod squares in the
```
heanlan, line 118: label -> Pod labels
pod -> Pod

plugins/grafana-custom-plugins/grafana-dependency-plugin/src/DependencyPanel.tsx
```diff
@@ -44,26 +61,45 @@ export const DependencyPanel: React.FC<Props> = ({ options, data, width, height
...
+        return source ? sourcePodName : destinationPodName;
+      }
+      let labels = JSON.parse(labelJSON);
+      if(labels[options.labelName] !== undefined) {
+        return labels[options.labelName];
+      }
+      return sourcePodName;
+    }
+
+    // TODO add button or switch to set groupByLabelVariable
```
heanlan, line 83: little confused on the TODO comment here. What is it for? I thought we already have a switch in the panel options?

heanlan (commented):
Could you please also update network-flow-visibility.md to add the introduction of the new options?

plugins/grafana-custom-plugins/grafana-dependency-plugin/src/module.ts
```diff
@@ -2,4 +2,38 @@ import { PanelPlugin } from '@grafana/data';
 import { DependencyOptions } from './types';
 import { DependencyPanel } from './DependencyPanel';
 
-export const plugin = new PanelPlugin<DependencyOptions>(DependencyPanel).setPanelOptions((builder) => builder);
+export const plugin = new PanelPlugin<DependencyOptions>(DependencyPanel).setPanelOptions((builder) => builder.addBooleanSwitch({
+    path: 'groupByLabel',
+    name: 'Group By Label',
```
heanlan, line 7: I think it would be better to change to Group by Pod Label

plugins/grafana-custom-plugins/grafana-dependency-plugin/src/DependencyPanel.tsx
Dhruv-J: I meant to delete that earlier

plugins/grafana-custom-plugins/grafana-dependency-plugin/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.2 - 04-xx-2023
+
+Added label view toggle and color choice.
```
heanlan, line 5: Added "Group by Pod Labels" toggle and Pod square color choices.

Dhruv-J:
> Could you please also update network-flow-visibility.md to add the introduction of the new options?

I left it off because it was part of the dependency plugin instead, so I added it to the README for that. I can also add it to flow-visibility.md if it would fit there.

heanlan:
> I left it off because it was part of the dependency plugin instead, so I added it to the README for that. I can also add it to flow-visibility.md if it would fit there.

Eventually it is a user option, right? Normally users would only read network-flow-visibility.md but not plugin README. I think it would be good to let users know they have these options.

Dhruv-J:
> > I left it off because it was part of the dependency plugin instead, so I added it to the README for that. I can also add it to flow-visibility.md if it would fit there.
> 
> Eventually it is a user option, right? Normally Theia users would only read network-flow-visibility.md but not plugin README. I think it would be good to let users know they have these options.

Yeah, that makes sense, sounds good.

docs/network-flow-visibility.md
```diff
@@ -764,6 +764,12 @@ the selected time range.
 
 <img src="https://downloads.antrea.io/static/04132023/flow-visibility-network-topology-0.png" width="400" alt="Network Topology Dashboard service dependency graph">
 
+Users can group Pods by label, allowing Pod squares in the diagram to represent
+a set of pods with the same label value. It is also possible to choose the
```
heanlan, line 768:
```suggestion
a set of Pods with the same label value. It is also possible to choose the
```

docs/network-flow-visibility.md
```diff
@@ -764,6 +764,12 @@ the selected time range.
 
 <img src="https://downloads.antrea.io/static/04132023/flow-visibility-network-topology-0.png" width="400" alt="Network Topology Dashboard service dependency graph">
 
+Users can group Pods by label, allowing Pod squares in the diagram to represent
```
heanlan, line 767: maybe add some instructions to let users know where should they find the configurations, telling users it is in panel edit

build/charts/theia/README.md
```diff
@@ -52,7 +52,7 @@ Kubernetes: `>= 1.16.0-0`
...
 | grafana.image | object | `{"pullPolicy":"IfNotPresent","repository":"projects.registry.vmware.com/antrea/theia-grafana","tag":"8.3.3"}` | Container image used by Grafana. |
-| grafana.installPlugins | list | `["https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.2.zip;theia-grafana-sankey-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-chord-plugin-1.0.1.zip;theia-grafana-chord-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-dependency-plugin-1.0.1.zip;theia-grafana-dependency-plugin","grafana-clickhouse-datasource 1.0.1"]` | Grafana plugins to install. |
+| grafana.installPlugins | list | `["https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.2.zip;theia-grafana-sankey-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-chord-plugin-1.0.1.zip;theia-grafana-chord-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-chord-plugin-1.0.2.zip;theia-grafana-dependency-plugin","grafana-clickhouse-datasource 1.0.1"]` | Grafana plugins to install. |
```
heanlan, line 55: chord -> dependency

plugins/grafana-custom-plugins/grafana-dependency-plugin/README.md
```diff
@@ -42,18 +42,24 @@ loading of data for the Service Dependency Graph Plugin, the query is expected
 to return the following fields, in arbitrary order.
 
 - field 1: sourcePodName value with name or an alias of `sourcePodName`
-- field 2: sourceNodeName value with name or an alias of `sourceNodeName`
-- field 3: destinationPodName value with name or an alias of `destinationPodName`
-- field 4: destinationNodeName value with name or an alias of `destinationNodeName`
-- field 5: destinationServicePortName value with name or an alias of `destinationServicePortName`
-- field 6: octetDeltaCount value with name or an alias of `octetDeltaCount`
+- field 2: sourcePodLabels value with name or alias of `sourcePodLabels`
+- field 3: sourcePodNamespace value with name or alias of `sourcePodNamespace`
```
heanlan, line 46: why we add `sourcePodNamespace` here? Is it being used anywhere in the source code?

heanlan (approved):
LGTM

Issue #163, opened 2023-02-08
Improve Code Coverage 


elton-furtado:
Current code coverage overall in Theia is 50%, we need to improve the coverage to at least 60% in the next release.

There is also codecov bot reporting missing in every PR, this needs to be investigated further and resolved

PR #230, opened 2023-04-04
Add missing Unit Tests

tushartathgur:
For #163

### DIFF
--- a/pkg/theia/commands/anomaly_detection_run_test.go
+++ b/pkg/theia/commands/anomaly_detection_run_test.go
@@ -16,6 +16,7 @@ package commands
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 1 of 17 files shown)


pkg/theia/commands/anomaly_detection_run_test.go
```diff
@@ -89,15 +127,18 @@ func TestAnomalyDetectionRun(t *testing.T) {
 			cmd.Flags().Bool("use-cluster-ip", true, "")
 			cmd.Flags().String("algo", "ARIMA", "")
 			cmd.Flags().String("start-time", "2006-01-02 15:04:05", "")
-			cmd.Flags().String("end-time", "2006-01-03 15:04:05", "")
+			cmd.Flags().String("end-time", "2006-01-03 16:04:05", "")
```
elton-furtado, line 130: whats the reason for extending by 1 hour?

pkg/theia/commands/policy_recommendation_retrieve_test.go
```diff
@@ -95,23 +96,74 @@ func TestPolicyRecommendationRetrieve(t *testing.T) {
 			expectedMsg:      []string{},
 			expectedErrorMsg: "error when getting policy recommendation job",
 		},
+		{
+			name:             "Unspecified name",
+			testServer:       httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})),
+			nprName:          "pr-e292395c-3de1-11ed-b878-0242ac120001",
```
elton-furtado, line 102: we shouldn't hardcode name and use random ID all over the test cases

pkg/theia/commands/anomaly_detection_run_test.go
tushartathgur: because the end time should be after the start-time,

pkg/theia/commands/policy_recommendation_retrieve_test.go
tushartathgur: we are testing it for the cases when we have the job id, we have to use this job id in order to retrieve

pkg/theia/commands/anomaly_detection_run_test.go
elton-furtado: If you know the execution time for all the times, lets set that value instead of waiting for such a long time

pkg/theia/commands/policy_recommendation_retrieve_test.go
elton-furtado: you can randomly create job id in every test case, instead of hardcoding it in all the files

pkg/theia/commands/anomaly_detection_run_test.go
tushartathgur: This is a table entry, it has nothing to do with how long the job will run for

pkg/theia/commands/policy_recommendation_run_test.go
```diff
@@ -143,3 +156,355 @@ func TestPolicyRecommendationRun(t *testing.T) {
...
+			cmd.Flags().String("ns-allow-list", "[\"kube-system\",\"flow-aggregator\",\"flow-visibility\"]", "")
+			cmd.Flags().Bool("exclude-labels", true, "")
+			cmd.Flags().Bool("to-services", true, "")
+			cmd.Flags().Int32("executor-instances", 1, "")
+			cmd.Flags().String("driver-core-request", "1", "")
+			cmd.Flags().String("driver-memory", "1m", "")
+			cmd.Flags().String("executor-core-request", "1", "")
+			cmd.Flags().String("executor-memory", "1m", "")
+			cmd.Flags().String("file", "filename", "")
+			cmd.Flags().Bool("use-cluster-ip", true, "")
```
dreamtalen, line 499: A Similiar case here, consider my suggestion to deduplicate code from line 277 to 499.

pkg/theia/commands/anomaly_detection_run_test.go
```diff
@@ -107,3 +148,215 @@ func TestAnomalyDetectionRun(t *testing.T) {
...
+		case "Unspecified use-cluster-ip":
+			cmd.Flags().String("algo", "ARIMA", "")
+			cmd.Flags().String("start-time", "2006-01-02 15:04:05", "")
+			cmd.Flags().String("end-time", "2006-01-03 16:04:05", "")
+			cmd.Flags().String("ns-ignore-list", "[\"kube-system\",\"flow-aggregator\",\"flow-visibility\"]", "")
+			cmd.Flags().Int32("executor-instances", 1, "")
+			cmd.Flags().String("driver-core-request", "1", "")
+			cmd.Flags().String("driver-memory", "1m", "")
+			cmd.Flags().String("executor-core-request", "1", "")
+			cmd.Flags().String("executor-memory", "1m", "")
```
dreamtalen, line 352: Hi Tushar, I noticed that there are many duplicate flag settings code between lines 241-352. To improve this, one possible approach that comes to mind is to preset a cmd variable with all correct flags, and then in each test case, use `markHidden` to remove/hidden a flag for the "Unspecified xxxx" test cases and set the flag to a mock value for the "Invalid xxxx" test cases. This would make the code look cleaner and more organized in my opinion.

tushartathgur: currently we can not find an approach to reset the flags, resolving this conversation for now

pkg/theia/commands/policy_recommendation_run_test.go
tushartathgur: can find an approach for now, resolving the conversation

dreamtalen (approved):
LGTM, no further comments at this time.

PR #249, opened 2023-04-11
Combining Dependabot PRs workflow

Dhruv-J:
This PR adds a workflow to github that combines existing dependabot PRs. Workflow gotten from: https://github.com/hrvey/combine-prs-workflow .


Dhruv-J:
Don't think it is needed anymore, as vulnerabilities have severely decreased in both quantity and frequency of PR making.

PR #250, opened 2023-04-11
Re-organize CHANGELOG for automatic updates

tushartathgur:
We introduce a similar workflow like Antrea which will take care of automatically updating the appropriate CHANGELOG file on the main branch whenever a new version of Theia (minor or bug fix) is released. This will ensure that the CHANGELOG on the main branch is always up-to-date. The workflow cannot commit to the main branch directly (it is protected), but it will open a PR automatically.

### DIFF
--- a/.github/workflows/update_changelog.yml
+++ b/.github/workflows/update_changelog.yml
@@ -1,0 +1,60 @@
+name: Update CHANGELOG after release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  check-version:
+    runs-on: [ubuntu-latest]
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+    steps:
+    - name: Extract version from Github ref
+      id: get-version
+      env:
+        TAG: ${{ github.ref }}
+      shell: bash
+      run: |
+        version=${TAG:10}
+        if [[ "$version" == *-* ]]; then
+          echo "$version is a release candidate or a pre-release"
+          exit 0
+        fi
+        echo "::set-output name=version::$version"
+
+  pr-update-changelog:
+    runs-on: [ubuntu-latest]
+    needs: check-version
+    if: ${{ needs.check-version.outputs.version != '' }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: main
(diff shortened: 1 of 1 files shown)


dreamtalen:
As my knowledge about this GitHub workflow is limited, could you also request a review from Antonin?

antoninbas (commented):
The commit message doesn't seem right:

> This lets us have a different CHANGELOG file for each minor release train (same as K8s), which makes it easier to navigate changes...

Looking at the Theia repo, this is already the case. There is already a different CHANGELOG file for each minor release train.

tushartathgur:
> The commit message doesn't seem right:
> 
> > This lets us have a different CHANGELOG file for each minor release train (same as K8s), which makes it easier to navigate changes...
> 
> Looking at the Theia repo, this is already the case. There is already a different CHANGELOG file for each minor release train.

currently, we do it manually for Theia. we wanted to add an automatic way of updating and creating new CHANGELOG file

antoninbas:
> currently, we do it manually for Theia. we wanted to add an automatic way of updating and creating new CHANGELOG file

Then that's what the commit message should say. It seems that the current one was copied from the corresponding Antrea commit. For Antrea it made sense because when we made that change, there was a unique CHANGELOG file. For Theia, the message as it is doesn't make sense because we already have different CHANGELOG files. Only the second part of the commit message is appropriate.

tushartathgur:
> > currently, we do it manually for Theia. we wanted to add an automatic way of updating and creating new CHANGELOG file
> 
> Then that's what the commit message should say. It seems that the current one was copied from the corresponding Antrea commit. For Antrea it made sense because when we made that change, there was a unique CHANGELOG file. For Theia, the message as it is doesn't make sense because we already have different CHANGELOG files. Only the second part of the commit message is appropriate.



> > currently, we do it manually for Theia. we wanted to add an automatic way of updating and creating new CHANGELOG file
> 
> Then that's what the commit message should say. It seems that the current one was copied from the corresponding Antrea commit. For Antrea it made sense because when we made that change, there was a unique CHANGELOG file. For Theia, the message as it is doesn't make sense because we already have different CHANGELOG files. Only the second part of the commit message is appropriate.

That makes sense. Updated the commit message and description accordingly.

antoninbas (approved):

PR #266, opened 2023-04-18
Remove android library in action to save space storage for e2e test.

yuntanghsu:
Add `sudo rm -rf "/usr/local/lib/android"` to release ~10GB space for e2e test.

before
image

after
image

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -114,6 +114,7 @@ jobs:
           sudo rm -rf /opt/ghc
           sudo rm -rf "/usr/local/share/boost"
           sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+          sudo rm -rf "/usr/local/lib/android"
           df -h
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


heanlan (approved):
LGTM

I'm just curious about how did we decide to remove android instead of other libraries?

yuntanghsu:
> LGTM
> 
> I'm just curious about how did we decide to remove android instead of other libraries?

Those folders are pre-installed and users usually remove them in runtime to increase storage.
I think we won't use android library and it can provide ~12G space, so I delete it.

However, I think it is just a temporary workaround.
I haven't find the reason why our e2e test used all 41G space. (Our builded image only occupy 7G)

dreamtalen (approved):
LGTM.
Also noticed this script https://github.com/apache/flink/blob/master/tools/azure-pipelines/free_disk_space.sh Flink is using to free disk spaces. We can try using their commands if we run out of disk space again.

yanjunz97 (approved):

PR #269, opened 2023-04-24
Substituting the customized Clair scan, implement Trivy Github workflows

tushartathgur:
- To replace our outdated "custom" Clair scan code which didn't really help find vulnerabilities and Github workflow for running scans, we have included a Makefile target to perform Trivy scans.
- We will also leverage the standard aquasecurity/trivy-action Github action to execute new Github workflows.
- The file operates daily automatically or could be triggered manually by specifying image version and produces reports, which are then uploaded as artifacts.


elton-furtado:
Can you list the problems with the previous Clair scan?

.github/workflows/trivy_scan.yml
```diff
@@ -0,0 +1,163 @@
...
+    if: github.repository == 'antrea-io/theia'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Find greatest Antrea version
+      id: find-theia-greatest-version
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        VERSION=${{ github.event.inputs.theia-version }}
```
antoninbas, line 20: you are missing the `theia-version` input in the `workflow_dispatch` definition above

.github/workflows/trivy_scan.yml
```diff
@@ -0,0 +1,163 @@
...
+      # we cannot use .trivy.yml as we need to override some config parameters
+      # and that is not supported by aquasecurity/trivy-action
+      with:
+        scan-type: 'image'
+        image-ref: 'antrea/theia-spark-jobs:latest'
+        exit-code: '1'
+        ignore-unfixed: true
+        severity: 'CRITICAL,HIGH'
+        # whereabouts project doesn't upgrade dependencies frequently
+        skip-files: '/opt/cni/bin/whereabouts'
```
antoninbas, line 48: I doubt that whereabouts is included in any of the Theia images

.github/workflows/trivy_scan.yml
```diff
@@ -0,0 +1,163 @@
...
+        ignore-unfixed: true
+        severity: 'CRITICAL,HIGH'
+        skip-files: '/opt/cni/bin/whereabouts'
+        format: 'table'
+        output: 'trivy.clickhouse-monitor.latest.txt'
+    - name: Run Trivy vulnerability scanner on latest Theia clickhouse-server image
+      if: ${{ always() && steps.pull.conclusion == 'success' }}
+      uses: aquasecurity/trivy-action@0.10.0
+      # we cannot use .trivy.yml as we need to override some config parameters
+      # and that is not supported by aquasecurity/trivy-action
```
antoninbas, line 67: probably not useful to repeat this comment so many times

antoninbas (commented):
with many different images (4), you could consider using a job matrix: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

that would require moving the step used to determine the version to its own separate job.

.github/workflows/trivy_scan_before_release.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Build theia Docker images
+      id: make
+      run: |
+        make ${{ matrix.image }}
+    - name: Run Trivy vulnerability scanner on latest Theia ${{ matrix.image }} image
+      if: ${{ always() && steps.make.conclusion == 'success' }}
+      uses: aquasecurity/trivy-action@0.10.0
+      continue-on-error: true
+      # we cannot use .trivy.yml as we need to override some config parameters
+      # and that is not supported by aquasecurity/trivy-action
```
antoninbas, line 29: I think you meant to use that comment for the other workflow? For this one, you are actually using .trivy.yml for the config.

.github/workflows/trivy_scan_before_release.yml
```diff
@@ -0,0 +1,49 @@
...
+      with:
+        scan-type: 'image'
+        image-ref: 'antrea/theia-${{ matrix.image }}:latest'
+        trivy-config: '.trivy.yml'
+  build_theia_manager:
```
antoninbas, line 34: it's not obvious to me why `theia-manager` is not just another case in the previous job's matrix?

.github/workflows/trivy_scan_before_release.yml
```diff
@@ -0,0 +1,49 @@
...
+          - clickhouse-monitor
+          - clickhouse-server
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build theia Docker images
+      id: make
+      run: |
+        make ${{ matrix.image }}
+    - name: Run Trivy vulnerability scanner on latest Theia ${{ matrix.image }} image
+      if: ${{ always() && steps.make.conclusion == 'success' }}
```
antoninbas, line 25: it feels like this if condition doesn't serve much of a purpose? `always()` is used to make sure that the job always runs, even if previous steps failed. And you are adding a condition to only run the job is the previous step is successful. Let me know if I'm missing something.

.github/workflows/trivy_scan_before_release.yml
tushartathgur: I did it this way as for all other images they do not have “theia-“ prefix while manager does, so either I change the Strings of all the make cases or I add an additional job here, so I preferred doing the later and the code change is quite less as compared to the former.

antoninbas: Sounds good, that's what I was missing

.github/workflows/trivy_scan_before_release.yml
tushartathgur: In order to fetch the latest image, we should be relying for the make to pass in the prev step.

antoninbas: We are not fetching the image, I think we are building it locally?
My point is that if you omit this one line altogether, it should have the same effect (i.e., skip this step if the previous one failed).

tushartathgur: oh ok, yes that make sense

.github/workflows/trivy_scan.yml
```diff
@@ -0,0 +1,80 @@
...
+        continue-on-error: true
+        with:
+          scan-type: 'image'
+          image-ref: 'antrea/${{ matrix.image }}:${{ matrix.version }}'
+          exit-code: '1'
+          ignore-unfixed: true
+          severity: 'CRITICAL,HIGH'
+          format: 'table'
+          output: 'trivy.${{ matrix.image }}.${{ matrix.version }}.txt'
+  upload_trivy_scan_reports:
```
antoninbas, line 63: I missed this before: jobs are independent in Github workflows (unlike steps), so the trivy report files will not exist  when that job runs. You need to upload artifacts as part of the previous job.

antoninbas (commented):
it may be a good idea to test the workflow in your personal fork before we merge

.github/workflows/trivy_scan.yml
tushartathgur: yes, I kept the “needs” and “if” for the same in this job in order for it to wait for the build job to be completed successfully and then upload the artifacts.

antoninbas: This is not about correct ordering of jobs. Different jobs do not share a file system. So when this job executes, `trivy.*.txt` will not select any file.

Different jobs execute in different VMs.

.github/workflows/trivy_scan.yml
```diff
@@ -0,0 +1,75 @@
...
+        continue-on-error: true
+        with:
+          scan-type: 'image'
+          image-ref: 'antrea/${{ matrix.image }}:${{ matrix.version }}'
+          exit-code: '1'
+          ignore-unfixed: true
+          severity: 'CRITICAL,HIGH'
+          format: 'table'
+          output: 'trivy.${{ matrix.image }}.${{ matrix.version }}.txt'
+      - name: upload_trivy_scan_reports
```
antoninbas, line 63: nit: we usually use a more "reader-friendly" name here, as it doesn't need to be a single word
There is also a single report at this stage (reports -> report)

antoninbas (approved):
LGTM

Issue #162, opened 2023-02-08
Add an operator for ClickHouse uninstallation


yanjunz97:
The ClickHouse cluster needs ZooKeeper for deletion. But when uninstalling Theia with yaml file or Helm chart as a whole, Zookeeper is always deleted before ClickHouse. This will lead to a fact that ClickHouse deletion will be stuck, while the ClickHouse Operator will keep trying to connect ZooKeeper for around 4 minutes and eventually fail.

Currently we instruct the user in [doc](https://github.com/antrea-io/theia/blob/main/docs/network-flow-visibility.md#clickhouse-cluster) to manually delete the ClickHouse first and then uninstall the rest of Theia. But for usability, we need an operator to take care of the installation and uninstallation order of ClickHouse cluster and ZooKeeper.

tushartathgur:
In Theia, we are leveraging ClickHouse-Keeper which will help us be on track with the new developments from ClickHouse.
we shall also be adding an upgrade script that helps users to upgrade seamlessly from zookeeper to ClickHouse-keeper without losing the data.

PR #281, opened 2023-05-01
Clickhouse-Keeper Support for Theia

tushartathgur:
- Zookeeper is an external entity when it comes to clickhouse and is maintained externally.
- Adding Clickhouse-keeper as it is developed by clickhouse and is an alternate to Zookeeper.

#162


hack/clickhouse-backup-restore/main.go
```diff
@@ -0,0 +1,43 @@
...
+
+	fmt.Println("Cluster update from:", *fromYaml)
+	fmt.Println("Cluster update to:", *toYaml)
+
+	// Check if to and from yamls are provided.
+	if *toYaml == "" || *fromYaml == "" {
+		fmt.Println("Please provide both YAML files for upgrade ")
+		os.Exit(1)
+	}
+	clickhouseBackupRestore.BackupAndRestoreClickhouse(*fromYaml, *toYaml, *deleteBackup, *loglevel, *deleteYamls)
```
yanjunz97, line 42: I haven't looked into the details, but generally I have the following considerations

- I notice you are using a third party open source code from https://github.com/AlexAkulov/clickhouse-backup. I'm not quite sure about whether there will be a license issue as they are using an MIT License while we are using the Apache License. cc @elton-furtado for suggestions
- Can we use shell instead of golang? Otherwise users have to install golang and compile it locally only for ClickHouse upgrading/downgrading. If the logic is too complex to be implemented in shell script, maybe we can think about moving it to theia-cli or adding it as a separate binary in release.
- The script expects users to install Theia with Yaml. But it might not work well for users using Helm to do version control on Theia. Can we also add options for Helm chart upgrading?

elton-furtado: we will have to add MIT license in our documentation, if we ever add this support

tushartathgur:
Reason for Closing the commit: 
The update script doesn't work as expected as before because upgrade script needs to remove zookeeper and use clickhouse-keeper while maintaining the persistent volume, else we lose the data

Issue #285, opened 2023-05-04
Coverage for Python files


Dhruv-J:
Currently, python files, specifically spark jobs, are not described through code coverage. The setup required to generate the python coverage already exists, Codecov just needs to point to the generated file.

PR #296, opened 2023-05-12
Python Coverage

Dhruv-J:
Adds python coverage to codecov. Infrastructure already exists, changes made allow coverage information to be viewed on codecov website and PR comments.

Fixes issue #285


Dhruv-J:
One note: Although the flag for python shows up in the comment and analysis of codecov, the report generated and uploaded has a name of just numbers:
Screenshot 2023-05-15 at 4 59 30 PM
I tried seeing if there was a setting or additional argument I could give in order to change this name, but I couldn't find much.

elton-furtado (approved):

PR #298, opened 2023-05-16
Coverage Target and Thresholds

Dhruv-J:
Adds target for patch, as well as thresholds for project and unit-tests flag.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -18,7 +18,7 @@ coverage:
   status:
     patch:
       default:
-        target: auto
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -18,7 +18,7 @@ coverage:
   status:
     patch:
       default:
-        target: auto
+        target: 60%
```
elton-furtado, line 21: make it 69%

elton-furtado (approved):

PR #310, opened 2023-05-25
Upload ClickHouse-Keeper image in Theia.

tushartathgur:
Adding ClickHouse-Keeper image for kind and unit tests and for manifest yamls

### DIFF
--- a/.github/workflows/docker_update_theia.yml
+++ b/.github/workflows/docker_update_theia.yml
@@ -19,6 +19,9 @@ on:
       zookeeper-tag:
         description: 'Tag for zookeeper Docker image'
         required: true
+      clickhouse-keeper-tag:
+        description: 'Tag for ClickHouse-keeper Docker image'
+        required: true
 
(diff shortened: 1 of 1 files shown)


elton-furtado:
let's keep all the clickhouse-keeper in one PR

tushartathgur:
> let's keep all the clickhouse-keeper changes in one PR

we will have to keep this update separate for the test cases to be able to fetch the image from the repository, we will first have to upload the image to the repository.

elton-furtado:
> 

which testcases?  for now we shouldn't be adding any new tests for clickhouse-keeper.

tushartathgur:
> which testcases? for now we shouldn't be adding any new tests for clickhouse-keeper.

in the PR #281, we were using click house-keeper image directly from docker, which instead should be first uploaded to the Theia image repo in harbor and should then be pulled for using in test cases, test cases currently in #281 would fail if the image is not present.

PR #317, opened 2023-06-02
Fix flowvisibility e2e test by enable FlowExporter toggle

heanlan:
A configuration toggle for flow exporter has recently been added to antrea agent config. We need to enable both the feature gate and the toggle to enable flow exporter feature.

https://github.com/antrea-io/antrea/pull/5021


heanlan:
jenkins-e2e are failing for other reasons, need more triage. I think we can merge this PR for making kind e2e back to normal first.

yuntanghsu (approved):
LGTM

PR #348, opened 2023-06-21
Add ClickHouse server TLS option

yanjunz97:
Generally, we recommend using Ingress/LoadBalancer for secure connection to ClickHouse.
But we also provide this TLS option at ClickHouse server side.

Currently ClickHouse Operator does not support cert-manager, which means users need to
manually do the cert rotation when using this ClickHouse server side TLS option.

~~This PR also includes a script to generate all required cert files for ClickHouse.
You could use it with the following command. It will generate the dh parameter, a server
cert, a server key and a CA cert in `build/charts/theia/provisioning/tls`.
You can provision the CA cert at the client side.~~

This PR also offers an option to use the certs generated by a functionality in helm.

### DIFF
--- a/build/charts/theia/templates/clickhouse/clickhouseinstallation.yaml
+++ b/build/charts/theia/templates/clickhouse/clickhouseinstallation.yaml
@@ -16,8 +16,18 @@ spec:
       {{ .Values.clickhouse.connectionSecret.readOnlyUsername }}/networks/ip: "::/0"
     profiles:
       readonly/readonly: 1
+    {{- if .Values.clickhouse.service.secureConnection.enable }}
+    files:
+      {{- include "clickhouse.tlsConfig" (dict "Files" .Files "Global" .) | indent 6 }}
+    {{- end }}
     clusters:
       - name: "clickhouse"
+        {{- if .Values.clickhouse.service.secureConnection.enable }}
+        settings:
+          tcp_port_secure: {{ .Values.clickhouse.service.secureConnection.secureTcpPort }}
+          https_port: {{ .Values.clickhouse.service.secureConnection.httpsPort }}
+        secure: "yes"
+        {{- end }} 
         layout:
           {{- if gt (int .Values.clickhouse.cluster.replicas) 1 }}
           shards:
@@ -66,8 +76,16 @@ spec:
           ports:
             - name: http
               port: {{ .Values.clickhouse.service.httpPort }}
+              targetPort: 8123
             - name: tcp
               port: {{ .Values.clickhouse.service.tcpPort }}
+              targetPort: 9000
+            {{- if .Values.clickhouse.service.secureConnection.enable }}
+            - name: https
+              port: {{ .Values.clickhouse.service.secureConnection.httpsPort }}
+            - name: secureclient
+              port: {{ .Values.clickhouse.service.secureConnection.secureTcpPort }}
+            {{- end }}
     podTemplates:
       {{- if gt (int .Values.clickhouse.cluster.replicas) 1 }}
       - name: pod-template-without-monitor
--- a/build/charts/theia/values.yaml
+++ b/build/charts/theia/values.yaml
@@ -43,10 +43,37 @@ clickhouse:
     # -- The type of Service exposing ClickHouse. It can be one of ClusterIP,
     # NodePort or LoadBalancer.
     type: ClusterIP
-    # -- TCP port number for the ClickHouse service.
+    # -- TCP port number for ClickHouse service.
     tcpPort: 9000
-    # -- HTTP port number for the ClickHouse service.
+    # -- HTTP port number for ClickHouse service.
     httpPort: 8123
+    # As ClickHouse Operator currently does not support the cert management,
+    # we recommend to use Ingress or Load Balancer to expose ClickHouse securely.
+    # But we also provide this option to provide the secure connection from
+    # ClickHouse server end. To use this option, please rotate the certs manually.
+    secureConnection:
+      # -- Determine whether to enable secure connection to ClickHouse
+      enable: false
+      # -- TCP protocol TLS port.
+      secureTcpPort: 9440
+      # -- HTTP protocol TLS port.
+      httpsPort: 8443
+      # -- Indicates whether to use auto-generated self-signed TLS certificates.
+      # If false, a Secret named "clickhouse-tls" must be provided with the
+      # following keys: tls.crt and tls.key.
+      # If true, the following fields commonName, ipAddresses, dnsNames, daysValid
+      # need to be provided.
+      selfSignedCert: true
+      # -- Subject's common name. Only used when selfSignedCert is true.
+      commonName: "clickhouse-clickhouse.flow-visibility.svc"
+      # -- Optional list of IPs; may be nil. Only used when selfSignedCert is true.
+      ipAddresses: []
+      # -- Optional list of alternate DNS names; may be nil. Only used when
+      # selfSignedCert is true.
+      dnsNames: []
+      # -- Number of days for which the certificate will be valid. There is no automatic
+      # rotation with this method. This is ignored if selfSignedCert is false.
+      daysValid: 365
   # -- Time to live for data in the ClickHouse. Can be a plain integer using
   # one of these unit suffixes SECOND, MINUTE, HOUR, DAY, WEEK, MONTH, QUARTER,
   # YEAR.
--- a/docs/network-flow-visibility.md
+++ b/docs/network-flow-visibility.md
@@ -11,6 +11,7 @@
   - [Configuration](#configuration)
     - [With Helm](#with-helm)
       - [ClickHouse Cluster](#clickhouse-cluster)
+      - [Secure Connection](#secure-connection)
     - [With Standalone Manifest](#with-standalone-manifest)
       - [Grafana Configuration](#grafana-configuration)
         - [Service Customization](#service-customization)
@@ -288,6 +289,23 @@ ClickHouse cluster can be deployed with default Local PV or NFS PV by setting
 `clickhouse.storage.createPersistentVolume`. To have more flexibility in the
 PV creation, you can configure a customized `StorageClass` in
 `clickhouse.storage.persistentVolumeClaimSpec`.
+
+##### Secure Connection
+
+From Theia 0.8, we support TLS connection option in ClickHouse server. To
+enable TLS connection, please set `clickhouse.service.secureConnection.enable`
+to true.
+
+Currently ClickHouse Operator does not support cert-manager, which means users
+need to manually do the certs rotation when using this ClickHouse server TLS
+option. You can set `clickhouse.service.secureConnection.selfSignedCert` to
+true to use the default certs. Please ensure that you have set the subject's
+common name. You will need to renew the certs after `clickhouse.service.secureConnection.daysValid`
+passes.
+
+You can set `clickhouse.service.secureConnection.selfSignedCert` to
+false to provide your own certificates by creating a Secret with name
+`clickhouse-tls` containing the following keys: `tls.crt` and `tls.key`.
 
 #### With Standalone Manifest
 
--- a/build/charts/theia/README.md
+++ b/build/charts/theia/README.md
@@ -37,8 +37,16 @@ Kubernetes: `>= 1.16.0-0`
 | clickhouse.monitor.image | object | `{"pullPolicy":"IfNotPresent","repository":"projects.registry.vmware.com/antrea/theia-clickhouse-monitor","tag":""}` | Container image used by the ClickHouse Monitor. |
 | clickhouse.monitor.skipRoundsNum | int | `3` | The number of rounds for the monitor to stop after a deletion to wait for the ClickHouse MergeTree Engine to release memory. |
 | clickhouse.monitor.threshold | float | `0.5` | The storage percentage at which the monitor starts to delete old records. Vary from 0 to 1. |
-| clickhouse.service.httpPort | int | `8123` | HTTP port number for the ClickHouse service. |
-| clickhouse.service.tcpPort | int | `9000` | TCP port number for the ClickHouse service. |
+| clickhouse.service.httpPort | int | `8123` | HTTP port number for ClickHouse service. |
+| clickhouse.service.secureConnection.commonName | string | `"clickhouse-clickhouse.flow-visibility.svc"` | Subject's common name. Only used when selfSignedCert is true. |
+| clickhouse.service.secureConnection.daysValid | int | `365` | Number of days for which the certificate will be valid. There is no automatic rotation with this method. This is ignored if selfSignedCert is false. |
+| clickhouse.service.secureConnection.dnsNames | list | `[]` | Optional list of alternate DNS names; may be nil. Only used when selfSignedCert is true. |
+| clickhouse.service.secureConnection.enable | bool | `false` | Determine whether to enable secure connection to ClickHouse |
+| clickhouse.service.secureConnection.httpsPort | int | `8443` | HTTP protocol TLS port. |
+| clickhouse.service.secureConnection.ipAddresses | list | `[]` | Optional list of IPs; may be nil. Only used when selfSignedCert is true. |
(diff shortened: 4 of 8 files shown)


yanjunz97:
This PR is still in draft as we haven't adopt the secure connection support at the Client side.
I've only tested it with the local connection(without k8s).

Here is a reference to test this ClickHouse TLS support before we add support on the Client side.

1. Use this command to generate the certs

```shell
./hack/generate-clickhouse-cert.sh --SAN IP:127.0.0.1 --dhparam
```

2. After setting `clickhouse.service.secureConnection.enable` to true and deploying the ClickHouse,
you can use the following command to expose the secure port.

```shell
kubectl port-forward svc/clickhouse-clickhouse -n flow-visibility 9440:9440
```

3. You can use the following code to connect to the ClickHouse through the secure TCP port.
Make sure `ca-cert.pem` is reachable and has the same contents as the one generated in
`build/charts/theia/provisioning/tls`.

```golang
        caCert, err := os.ReadFile("ca-cert.pem")
	if err != nil {
		klog.Fatal(err)
	}
	caCertPool := x509.NewCertPool()
	successful := caCertPool.AppendCertsFromPEM(caCert)
	if !successful {
		klog.Fatal(err)
	}

	opt := clickhouse.Options{
		Addr: []string{"127.0.0.1:9440"},
		Auth: clickhouse.Auth{
			Database: "default",
			Username: "clickhouse_operator",
			Password: "clickhouse_operator_password",
		},
		TLS: &tls.Config{
			InsecureSkipVerify: false,
			RootCAs:            caCertPool,
		},
	}

	connect := clickhouse.OpenDB(&opt)
```

yanjunz97:
This PR is updated referring to: https://github.com/antrea-io/antrea/pull/5171. It adapts the self-generated cert by a functionality in helm, and removes the scripts in hack to generate the certs.

build/charts/theia/templates/clickhouse/clickhouseinstallation.yaml
```diff
@@ -16,8 +16,20 @@ spec:
...
+    {{- if .Values.clickhouse.service.secureConnection.enable }}
+    files:
+      {{- include "clickhouse.tlsConfig" (dict "Files" .Files "Global" .) | indent 6 }}
+    {{- end }}
     clusters:
       - name: "clickhouse"
+        {{- if .Values.clickhouse.service.secureConnection.enable }}
+        secure: "yes"
+        settings:
+          tcp_port: {{ .Values.clickhouse.service.tcpPort }} # keep for localhost
```
yuntanghsu, line 28: Do we want to expose tcp_port and http_port when secureConnection is disabled?
I only enclose https_port and tcp_port_secure inisde the double curly braket.
I'm not quite familiar with the settings, are they optional?

yanjunz97: Thanks for pointing out this! I've moved it out.

Let me explain about how it works.

The port here defines for the port in ClickHouse pod while the port exposed by the service defined in the Service template: https://github.com/antrea-io/theia/blob/aaeab26bb8c368ab0f243ded087148d73d280e41/build/charts/theia/templates/clickhouse/clickhouseinstallation.yaml#L66-L70

Although we may prefer to get consistency between these two, ClickHouse itself has some internal calls on the tcp port 9000 before the ClickHouse server is ready for the ClickHouse service, i.e. 9000 is hard-coded in ClickHouse code and I do not find a way to overwrite them easily. Thus finally my solution is to make the internal tcp port always 9000 while others will be consistent with the customization. And this will not affect the user experience.

yuntanghsu (approved):
LGTM

build/charts/theia/values.yaml
```diff
@@ -47,6 +47,33 @@ clickhouse:
...
     # -- HTTP port number for the ClickHouse service.
     httpPort: 8123
+    # As ClickHouse Operator currently does not support the cert management,
+    # we recommend to use Ingress or Load Balancer to expose ClickHouse securely.
+    # But we also provide this option to provide the secure connection from
+    # ClickHouse server end. To use this option, please rotate the certs manually.
+    secureConnection:
+      # -- Determine whether to enable secure connection to ClickHouse
+      enable: false
+      # -- TCP port number for the ClickHouse service.
```
heanlan, line 57:
```suggestion
      # -- TCP protocol TLS port.
```

build/charts/theia/values.yaml
```diff
@@ -47,6 +47,33 @@ clickhouse:
...
+      secureTcpPort: 9440
+      # -- HTTP port number for the ClickHouse service.
```
heanlan, line 59:
```suggestion
      # -- HTTP protocol TLS port.
```

build/charts/theia/values.yaml
```diff
@@ -47,6 +47,33 @@ clickhouse:
     tcpPort: 9000
     # -- HTTP port number for the ClickHouse service.
```
heanlan, line 48:
```suggestion
    # -- HTTP port number for ClickHouse service.
```

build/charts/theia/values.yaml
```diff
@@ -47,6 +47,33 @@ clickhouse:
...
+      enable: false
+      # -- TCP port number for the ClickHouse service.
+      secureTcpPort: 9440
+      # -- HTTP port number for the ClickHouse service.
+      httpsPort: 8443
+      # -- Indicates whether to use auto-generated self-signed TLS certificates.
+      # If false, a Secret named "clickhouse-tls" must be provided with the
+      # following keys: tls.crt and tls.key.
+      selfSignedCert: true
+      # -- Common name to use in the certificate. This is ignored if selfSignedCert
```
heanlan, line 65: I feel it is better to avoid double negatives. Same for below.
```suggestion
      # -- Subject's common name. Only used when selfSignedCert is true.
```

build/charts/theia/values.yaml
```diff
@@ -47,6 +47,33 @@ clickhouse:
...
+      # is false.
+      commonName: "clickhouse-clickhouse.flow-visibility.svc"
+      # -- IP addresses to use in the certificate. This is ignored if selfSignedCert
```
heanlan, line 68:
```suggestion
      # -- Optional list of IPs; may be nil. Only used when selfSignedCert is true.
```

build/charts/theia/values.yaml
```diff
@@ -47,6 +47,33 @@ clickhouse:
...
+    secureConnection:
+      # -- Determine whether to enable secure connection to ClickHouse
+      enable: false
+      # -- TCP port number for the ClickHouse service.
+      secureTcpPort: 9440
+      # -- HTTP port number for the ClickHouse service.
+      httpsPort: 8443
+      # -- Indicates whether to use auto-generated self-signed TLS certificates.
+      # If false, a Secret named "clickhouse-tls" must be provided with the
+      # following keys: tls.crt and tls.key.
```
heanlan, line 63: Maybe add "If true, the following fields commonName, ipAddresses, dnsNames, daysValid need to be provided."?

heanlan (commented):
I don't have strong opinion on refactoring the description of commonName, ipAddresses, dnsNames these three fields. But I suggest to specify ipAddresses, dnsNames are optional.

heanlan:
I notice we didn't include documentation update in this PR. @yuntanghsu will open a PR for adding documentation of supporting TLS connection between FA and CH. Do we plan to add it together in @yuntanghsu's PR later?

yuntanghsu:
> I notice we didn't include documentation update in this PR. @yuntanghsu will open a PR for adding documentation of supporting TLS connection between FA and CH. Do we plan to add it together in Yun-Tang's PR?

For the multi-cluster document, as we haven't finished the story of multi-cluster, Elton is thinking to add that document later.

heanlan:
> For the multi-cluster document, as we haven't finished the story of multi-cluster, Elton is thinking to add that document later.

I suppose we can add documentation on how to configure TLS connection between FA and CA?

build/charts/theia/values.yaml
yanjunz97: Thanks Anlan for reviewing! Updated.

yanjunz97:
> I suppose we can add documentation on how to configure TLS connection between FA and CH?

That makes sense. I added some doc for secure connection in the PR. Please take a look when you have time. Thanks!

docs/network-flow-visibility.md
```diff
@@ -289,6 +290,23 @@ ClickHouse cluster can be deployed with default Local PV or NFS PV by setting
...
+
+Currently ClickHouse Operator does not support cert-manager, which means users
+need to manually do the certs rotation when using this ClickHouse server TLS
+option. You can set `clickhouse.service.secureConnection.selfSignedCert` to
+true to use the default certs. Please ensure you set the Subject's common name
+and optional IP addresses and DNS names. You will need to renew the certs after
+`clickhouse.service.secureConnection.daysValid` passes.
+
+You can set `clickhouse.service.secureConnection.selfSignedCert` to
+false to use you own certicates by providing them in a Secret called
```
heanlan, line 307:
```suggestion
false to provide your own certificates by creating a Secret with name
```

docs/network-flow-visibility.md
```diff
@@ -289,6 +290,23 @@ ClickHouse cluster can be deployed with default Local PV or NFS PV by setting
...
+##### Secure Connection
+
+From Theia 0.7, we support TLS connection option in ClickHouse server. To
+enable TLS connection, please set `clickhouse.service.secureConnection.enable`
+to true.
+
+Currently ClickHouse Operator does not support cert-manager, which means users
+need to manually do the certs rotation when using this ClickHouse server TLS
+option. You can set `clickhouse.service.secureConnection.selfSignedCert` to
+true to use the default certs. Please ensure you set the Subject's common name
```
heanlan, line 302:
```suggestion
true to use the default certs. Please ensure that you have set the subject's common name
```
not sure, do we need to capitalize `subject`? And I think we can remove "and optional IP addresses and DNS names". It is a bit strange to me to say "ensure set ... optional ..."

test/e2e/framework.go
```diff
@@ -975,6 +975,21 @@ func (data *TestData) createBusyboxPodOnNode(name string, ns string, nodeName st
...
 }
 
+// getClickHouseOperator retrieves the name of the ClickHouse Operator Pod (clickhouse-operator-*) running on a specific Node.
+func (data *TestData) getClickHouseOperator() (*corev1.Pod, error) {
+	listOptions := metav1.ListOptions{
+		LabelSelector: "app=clickhouse-operator",
+	}
+	pods, err := data.clientset.CoreV1().Pods(kubeNamespace).List(context.TODO(), listOptions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list Flow Aggregator Pod: %v", err)
```
yuntanghsu, line 985:
```suggestion
		return nil, fmt.Errorf("failed to list ClickHouse Operatorr Pod: %v", err)
```

test/e2e/framework.go
```diff
@@ -1278,6 +1293,34 @@ func (data *TestData) deployFlowVisibilityCommon(chOperatorYML, flowVisibilityYM
 	if err != nil || rc != 0 {
 		return fmt.Errorf("error when deploying the ClickHouse Operator YML: %v\n is %s available on the control-plane Node?", err, chOperatorYML)
 	}
+	// Check for clickhouse operator pod running again for db connection establishment
```
yuntanghsu, line 1296: Maybe we can check the deployment directly? 
https://github.com/antrea-io/antrea/blob/9c32416d85a919f0ab087e5b564afe808249295a/test/e2e/framework.go#L1084

docs/network-flow-visibility.md

yanjunz97:
I noticed a strange behavior when running the ClickHouse migration e2e test in this PR.

When we update the ClickHouse Operator, it will schedule a task to renew the ClickHouse pod,
i.e. even if we do not apply a new flow-visibility yaml to update the ClickHouse Installation,
it will still restart the ClickHouse pod according to the old manifest.

If we update the ClickHouse Operator followed by the new Flow Visibility manifest, usually it
will be able to combine the two tasks and update the ClickHouse pod once to fulfill the new
Flow Visibility manifest.

However, if we have some changes in the `.spec.configuration.clusters.settings` for ClickHouseInstallation,
(that's what I tried in this PR, maybe changes in `.spec.configuration` will have the similar behavior)
the ClickHouse will not being able to combine the two tasks. Instead, it will run the first task
to renew the ClickHouse with old manifest and then the second task to update the ClickHouse
with new manifest.

Here is the error may happen in the e2e test. The root cause is in the downgrading tests.
As we have already applied the lower-version Flow visibility manifest while the ClickHouse
Operator is still applying the high-version Flow visibility manifest, it may find that some
migrators are missing in the configmap. This will lead to the first task of ClickHouseInstallation
stuck for a while(usually several minutes) and timeout.

```
          Type     Reason                        Age                    From               Message
          ----     ------                        ----                   ----               -------
          Warning  MultiplePodDisruptionBudgets  4m54s                  controllermanager  Pod "flow-visibility"/"chi-clickhouse-clickhouse-0-0-0" matches multiple PodDisruptionBudgets.  Chose "clickhouse-clickhouse" arbitrarily.
          Normal   Scheduled                     4m54s                  default-scheduler  Successfully assigned flow-visibility/chi-clickhouse-clickhouse-0-0-0 to kind-worker
          Warning  MultiplePodDisruptionBudgets  4m54s (x3 over 4m54s)  controllermanager  Pod "flow-visibility"/"chi-clickhouse-clickhouse-0-0-0" matches multiple PodDisruptionBudgets.  Chose "clickhouse" arbitrarily.
          Warning  FailedMount                   2m51s                  kubelet            Unable to attach or mount volumes: unmounted volumes=[clickhouse-configmap-volume], unattached volumes=[chi-clickhouse-deploy-confd-clickhouse-0-0 clickhouse-storage-template kube-api-access-4rf4q clickhouse-monitor-coverage clickhouse-configmap-volume chi-clickhouse-common-configd chi-clickhouse-common-usersd]: timed out waiting for the condition
          Warning  FailedMount                   48s                    kubelet            Unable to attach or mount volumes: unmounted volumes=[clickhouse-configmap-volume], unattached volumes=[clickhouse-storage-template kube-api-access-4rf4q clickhouse-monitor-coverage clickhouse-configmap-volume chi-clickhouse-common-configd chi-clickhouse-common-usersd chi-clickhouse-deploy-confd-clickhouse-0-0]: timed out waiting for the condition
          Warning  FailedMount                   44s (x10 over 4m53s)   kubelet            MountVolume.SetUp failed for volume "clickhouse-configmap-volume" : configmap references non-existent config key: 000005_0-6-0.down.sql
2023/07/26 01:25:18 Exporting test logs to '/home/runner/work/theia/theia/log/TestMigrate/beforeTeardown.Jul26-01-25-18'
2023/07/26 01:25:19 Error when running kubectl command on control-plane Node, cmd:kubectl -n flow-visibility logs --all-containers chi-clickhouse-clickhouse-0-0-0
stdout:
stderr:Error from server (BadRequest): container "clickhouse-monitor" in pod "chi-clickhouse-clickhouse-0-0-0" is waiting to start: ContainerCreating
```

This should not break users usage as after several tries it will start to apply the second task
and everything keeps going on. But it will break the e2e tests. I feel simply extending the
timeout is not a good choice (and it turns out extending the current 6-min-waiting for
ClickHouse pod to 9-min is still not enough). But I also do not come up with a better solution.
The only thought from my side is we may consider to apply the new Flow Visibility yaml
after the first task is finished. But that's a little bit difficult to track.

As the changes in `.spec.configuration.clusters.settings` is not a must have in this PR,
I removed that for now. And I log this problem here in case we may need to revisit it in
the future. I also would like to hear your opinions on this.

cc @heanlan @yuntanghsu @tushartathgur

test/e2e/framework.go
yanjunz97: Thanks Yuntang! I finally removed the changes in e2e tests and added a discussion below.

test/e2e/framework.go
yanjunz97: Thanks! That makes sense to me. But I found simply check the ClickHouse Operator is not enough and removed these parts. For more details, you can see my comments below

heanlan (approved):
LGTM

PR #410, opened 2023-07-28
Fix Policy Recommendation E2E test in Kind Cluster

dreamtalen:
Kind cluster introduced a new Pod local-path-provisioner recently which
causes Policy Recommendation E2E test because of recomming extra Network
Policies.
To fix test, we add the new Pod's namespace local-path-storage
into the argument when starting the recommendation job in Kind cluster.


yuntanghsu (approved):
LGTM

PR #413, opened 2023-08-02
upgrade antrea to v1.13.0

yuntanghsu:
- k8s.io/apimachinery/pkg/util/sets.String was deprecated, and replaced with sets.Set[string]. sets.String.List() was replaced with sets.List(sets.Set[string]). Similar changes happened on sets.Int32, sets.Int64
- k8s.io/apiserver/pkg/registry/rest.Storage interface added requirement on implementation of Destroy() method
- k8s.io/client-go/tools/remotecommand.Executor.Stream was deprecated, and replaced with StreamWithContext
- NetworkPolicy struct is updated from crdv1alpha1 to v1beta1
- ANPAppliedToSpec is renamed to ANNPAppliedToSpec
- "k8s.io/kubectl/pkg/cmd/cp" changed the way to take the parameter `args`. CopyOptions no longer takes the parameter `args` in `copyOptions.Run`.  User's defined `args` are used in copyOptions.Complete instead.


pkg/support/dump.go
```diff
@@ -182,11 +187,14 @@ func (d *managerDumper) DumpZookeeperLog(basedir string) error {
 func (d *managerDumper) copyFromPod(namespace, pod, srcDir, dstDir string) error {
 	ioStreams, _, out, errOut := genericclioptions.NewTestIOStreams()
 	copyOptions := cp.NewCopyOptions(ioStreams)
-	copyOptions.Clientset = d.clientSet
-	copyOptions.ClientConfig = d.restConfig
-	copyOptions.Namespace = namespace
-
-	if err := copyOptions.Run([]string{pod + ":" + srcDir, dstDir}); err != nil {
+	factory := util.NewFactory(newRestClientGetter(namespace, d.restConfig))
```
heanlan, line 190: out of curiosity, where are the changes of this file coming from?

heanlan (commented):
LGTM overall

I suggest we list out the notable changes in the PR description, as an example: https://github.com/antrea-io/antrea/pull/4935

pkg/support/dump.go
yuntanghsu: I think it is because k8s.io/kubectl is updated from v0.24.0 to v0.26.4. copyOptions also changed the way it takes args in v0.26.4. Previously, it took args when we called `copyOptions.Run(args)`. But the args can only be set in `copyOptions.Complete` now.  I can't find the usage of new copyOptions, so I followed its library to create `util.Factory` and `*cobra.Command`, which will be used in `copyOptions.Complete`.

pkg/support/dump.go
```diff
@@ -182,11 +187,14 @@ func (d *managerDumper) DumpZookeeperLog(basedir string) error {
...
+	if err := copyOptions.Complete(factory, cp.NewCmdCp(factory, ioStreams), []string{pod + ":" + srcDir, dstDir}); err != nil {
+		return fmt.Errorf("failed to complete: %v", err)
```
dreamtalen, line 192: nit: better to add some context about copy operation in this log, like `return fmt.Errorf("could not run copy operation: %v", err)` in ilne 198

pkg/support/dump.go
```diff
@@ -182,11 +187,14 @@ func (d *managerDumper) DumpZookeeperLog(basedir string) error {
...
+	}
+	if err := copyOptions.Validate(); err != nil {
+		return fmt.Errorf("failed to validate: %v", err)
```
dreamtalen, line 195: Same above

dreamtalen (approved):
Small nits only, LGTM.

pkg/support/dump.go
yuntanghsu: Fixed. Thanks!

pkg/support/dump.go
yuntanghsu: Fixed. Thanks!

heanlan (approved):
LGTM

PR #418, opened 2023-08-08
Broadcast ns_allow_list in recommendation spark job

dreamtalen:
In this PR, we changed the allow namespace list to a PySpark Broadcast
variable. Previously, this allow namespace list is a global variable in
the Python file, but the MapReduce function couldn't get the latest
value if it's changed by user defined allow namespaces. So we choose to
use Broadcast variable to share the latest allow namespace list to the
distributed running map functions.

### DIFF
--- a/plugins/policy-recommendation/policy_recommendation_job.py
+++ b/plugins/policy-recommendation/policy_recommendation_job.py
@@ -52,7 +52,11 @@ FLOW_TABLE_COLUMNS = [
     'flowType',
 ]
 
-NAMESPACE_ALLOW_LIST = ["kube-system", "flow-aggregator", "flow-visibility"]
+NAMESPACE_ALLOW_LIST = [
+    "kube-system",
+    "flow-aggregator",
+    "flow-visibility",
+]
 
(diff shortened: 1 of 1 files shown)


plugins/policy-recommendation/policy_recommendation_job.py
```diff
@@ -1168,7 +1196,7 @@ def main(argv):
             option,
             start_time,
             end_time,
-            ns_allow_list,
+            broadcast_ns_allow_list.value,
```
yuntanghsu, line 1199: Do we need to check broadcast_ns_allow_list is None? Or it must have field `value` after `spark.sparkContext.broadcast(NAMESPACE_ALLOW_LIST)` is called?

dreamtalen: Yep `broadcast_ns_allow_list` would not be None here so I skip the check.
For other distribued running map functions like `generate_k8s_np()`, I added the None check to unblock the unit tests.

yuntanghsu (approved):
LGTM

Issue #449, opened 2023-09-12
Dependabot in Snowflake Folder


Dhruv-J:
Currently, the snowflake folder dependencies are not automatically checked for vulnerabilities, nor are PRs made by Dependabot to fix any vulnerabilities. This is because Dependabot hasn't been setup for the snowflake folder yet.

PR #452, opened 2023-09-14
Dependabot Support for Snowflake Folder

Dhruv-J:
This PR adds dependabot support for the snowflake folder, as it has a separate go.mod and go.sum. Currently dependabot has trouble resolving indirect dependencies, which causes e2e tests to fail. Adding dependabot support will allow automatic updates for indirect dependencies and smoother running of e2e tests.

Fixes Issue #449


Dhruv-J:
For reviewers, I pushed this code to my fork's main to test the PR generation. You can check it out [here](https://github.com/Dhruv-J/theia/pulls).

PR #470, opened 2023-09-19
Add persistent volume option for zookeeper

### LINKED ISSUE #387: Fix ClickHouse crashing after ZooKeeper lost data
There is a problem reported by a user when deployment ClickHouse with PV and default ZooKeeper.

**How to reproduce the problem**

After deploy the ClickHouse with PV and default ZooKeeper, if the ZooKeeper pod crashes first and followed by the crashes of the ClickHouse pod, the bug will show up. e.g., if we run the following commands, the ClickHouse pod will keep crashing.

```
kubectl delete pod zookeeper-0 -n flow-visibility 
kubectl delete pod chi-clickhouse-clickhouse-0-0-0 -n flow-visibility 
```

We will be able to see from the ClickHouse pod log

```
...
2023.07.20 23:05:30.492951 [ 106 ] {} <Warning> default.flows_local (b0182a24-d467-4e9e-8f4a-908322270b64): No metadata in ZooKeeper for /clickhouse/tables/0/default/flows_local: table will be in readonly mode.
...
2023.07.20 23:05:31.086664 [ 52 ] {93600396-a3b1-42ca-acbb-5545f10eca29} <Error> TCPHandler: Code: 242. DB::Exception: Table is in readonly mode (replica path: /clickhouse/tables/0/default/flows_local/replicas/chi-clickhouse-clickhouse-0-0). (TABLE_IS_READ_ONLY), Stack trace (when copying this message, always include the lines below):

0. DB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int, bool) @ 0xb8a4c1a in /usr/bin/clickhouse
1. DB::Exception::Exception<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&>(int, fmt::v8::basic_format_string<char, fmt::v8::type_identity<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&>::type>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) @ 0xb90e6d8 in /usr/bin/clickhouse
2. DB::StorageReplicatedMergeTree::alter(DB::AlterCommands const&, std::__1::shared_ptr<DB::Context const>, std::__1::unique_lock<std::__1::timed_mutex>&) @ 0x16904fac in /usr/bin/clickhouse
3. DB::InterpreterAlterQuery::executeToTable(DB::ASTAlterQuery const&) @ 0x161a1fe8 in /usr/bin/clickhouse
4. DB::InterpreterAlterQuery::execute() @ 0x1619ff72 in /usr/bin/clickhouse
5. ? @ 0x1656ce56 in /usr/bin/clickhouse
6. DB::executeQuery(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::shared_ptr<DB::Context>, bool, DB::QueryProcessingStage::Enum) @ 0x1656a675 in /usr/bin/clickhouse
7. DB::TCPHandler::runImpl() @ 0x1714c8ea in /usr/bin/clickhouse
8. DB::TCPHandler::run() @ 0x1715f1d9 in /usr/bin/clickhouse
9. Poco::Net::TCPServerConnection::start() @ 0x19dc77f3 in /usr/bin/clickhouse
10. Poco::Net::TCPServerDispatcher::run() @ 0x19dc8b71 in /usr/bin/clickhouse
11. Poco::PooledThread::run() @ 0x19f79e3b in /usr/bin/clickhouse
12. Poco::ThreadImpl::runnableEntry(void*) @ 0x19f77540 in /usr/bin/clickhouse
13. ? @ 0x7fe5224f7609 in ?
14. __clone @ 0x7fe52241c133 in ?

Received exception from server (version 22.6.9):
Code: 242. DB::Exception: Received from 127.0.0.1:9000. DB::Exception: Table is in readonly mode (replica path: /clickhouse/tables/0/default/flows_local/replicas/chi-clickhouse-clickhouse-0-0). (TABLE_IS_READ_ONLY)
(query: ALTER TABLE flows_local MODIFY TTL timeInserted + INTERVAL 12 HOUR;)
```

**What we want to do**

It turns out the root cause is as the ZooKeeper does not support PV, the data in ZooKeeper will completely lost if it crashes. ClickHouse will convert the table to READONLY in this case.

Here we have several strategies we can consider:

1.  ClickHouse provides some strategies to recover from this condition:

- https://clickhouse.com/docs/en/sql-reference/statements/system#restore-replica
- https://kb.altinity.com/altinity-kb-setup-and-maintenance/altinity-kb-zookeeper/altinity-kb-recovering-from-complete-metadata-loss-in-zookeeper/

    If we need to use this strategy, we will need to investigate how to use them during the pod initialization.

2. Add PV option for ZooKeeper. This can be done refer to the way we add PV for ClickHouse and this example:https://github.com/Altinity/clickhouse-operator/blob/master/deploy/zookeeper/quick-start-persistent-volume/zookeeper-1-node.yaml

There might be a risk here if the data in ZooKeeper PV lost.

IMO, it would be good to add the ZooKeeper PV first. And we may also would like to investigate how to avoid the crashing if the data lost from ZooKeeper side so that users can still recover through ClickHouse strategies. Basically, this error may not appear if we remove the alter table clause for the TTL. We may need to investigate a better way to update the TTL.

yuntanghsu:
When deploying ClickHouse with Persistent Volumes (PV) and the default ZooKeeper, if the ZooKeeper pod crashes before the ClickHouse pod does, ClickHouse will automatically switch the table to a READONLY state. This sequential occurrence leads to the ClickHouse pod encountering repeated crashes, because the data stored in ZooKeeper is entirely lost in the event of its crash.

In this PR, we add persistent volume option for zookeeper for user when they want to enable PV for ClickHouse to prevent above issue.

resolves #387


build/charts/theia/values.yaml
```diff
@@ -131,6 +131,8 @@ clickhouse:
     # Specification to install a default ZooKeeper cluster. This is ignored
     # if zookeeperHosts are provided. 
     installZookeeper:
+      # -- Determine whether to enable persistent volume for zookeeper
+      createPersistentVolume: false
```
yanjunz97, line 135: How about using the same options in ch and grafana to provide PV support for local/nfs, and maybe also hostpath?

yanjunz97 (commented):
Thanks YunTang for taking care of this! Could you also add fix #387 in the description?

Have you tested the ZooKeeper with PV? I think you may also need to create PV and StorageClass for ZooKeeper like

- https://github.com/antrea-io/theia/blob/main/build/charts/theia/templates/clickhouse/local-persistentvolume.yaml
- https://github.com/antrea-io/theia/blob/main/build/charts/theia/templates/clickhouse/nfs-persistentvolume.yaml
- https://github.com/antrea-io/theia/blob/main/build/charts/theia/templates/clickhouse/storageclass.yaml

You can refer to the ClickHouse or Grafana implementation for PV.

build/charts/theia/values.yaml
yuntanghsu: I've tested the Zookeeper with PV. And it works fine when we have the scenario in the description. Two PVs will be created.
image
Do we still need to create PV and StorageClass?

yuntanghsu: Thank for reviewing! I've updated the options to match those we used for ClickHouse and Grafana.

build/charts/theia/values.yaml
```diff
@@ -297,3 +299,37 @@ theiaManager:
     tlsMinVersion: ""
   # -- Log verbosity switch for Theia Manager.
   logVerbosity: 0
+zookeeper:
```
yanjunz97, line 302: I'm thinking about putting this section under installZookeeper instead of as an individual section, as this section should only be useful when user enable installZookeeper.

build/charts/theia/values.yaml
yanjunz97, line 135: As mentioned above, if we move the ZooKeeper storage section here, I think it's not necessary to have this boolean createPersistentVolume. Instead we can read from storage.createPersistentVolume.type to determine whether PV creation is required, as what we do for ch and grafana

build/charts/theia/values.yaml
```diff
@@ -297,3 +299,37 @@ theiaManager:
...
+      # -- Type of PersistentVolume. Can be set to "Local" or "NFS". Please set
+      # this value to use a PersistentVolume created by Theia.
+      type: ""
+      local:
+        # -- The local path. Required when type is "Local".
+        path: "/data/zookeeper"
+        # -- Affinity for the Local PersistentVolume. By default it requires to
+        # label the Node used to store the Zookeeper data with
+        # "antrea.io/zookeeper-data-node=".
+        affinity: { }
```
yanjunz97, line 318: Maybe remove the space here to keep consistent with previous empty object?
```suggestion
        affinity: {}
```

build/charts/theia/values.yaml
```diff
@@ -297,3 +299,37 @@ theiaManager:
...
+    # -- Specification for PersistentVolumeClaim. This is ignored if
+    # createPersistentVolume.type is non-empty. To use a custom PersistentVolume,
+    # please set
+    # storageClassName: ""
+    # volumeName: "<my-pv>".
+    # To dynamically provision a PersistentVolume, please set
+    # storageClassName: "<my-storage-class>".
+    # Memory storage is used if both createPersistentVolume.type and
+    # persistentVolumeClaimSpec are empty.
+    persistentVolumeClaimSpec: { }
```
yanjunz97, line 333: ditto

build/charts/theia/values.yaml
yuntanghsu: I forgot to delete it. Will remove it later.

build/charts/theia/values.yaml
yuntanghsu: I've moved it under installZookeeper. Thanks.

build/charts/theia/values.yaml
yuntanghsu: Fixed.

build/charts/theia/values.yaml
yuntanghsu: Fixed. Thanks.

docs/network-flow-visibility.md
```diff
@@ -638,6 +653,79 @@ should be set to your storage size.
       name: clickhouse-storage-volume
     ```
 
+To deploy ZooKeeper with Persistent Volumes, please follow the steps outlined
```
yanjunz97, line 656: Maybe add "from Theia 0.8"?

docs/network-flow-visibility.md
```diff
@@ -286,10 +286,25 @@ will lead to a complete data loss. In that case, please refer to the
 to recover the data.
 
 ClickHouse cluster can be deployed with default Local PV or NFS PV by setting
-`clickhouse.storage.createPersistentVolume`. To have more flexibility in the
-PV creation, you can configure a customized `StorageClass` in
+`clickhouse.storage.createPersistentVolume`. To have more flexibility in the PV
+creation, you can configure a customized `StorageClass` in
 `clickhouse.storage.persistentVolumeClaimSpec`.
 
+We also offer the option of deploying Zookeeper with Persistent Volume. You can
```
yanjunz97, line 293: Ditto

docs/network-flow-visibility.md
```diff
@@ -286,10 +286,25 @@ will lead to a complete data loss. In that case, please refer to the
...
+configure a customized StorageClass within the
+`clickhouse.cluster.installZookeeper.storage.persistentVolumeClaimSpec`.
+
+It's important to select ZooKeeper with Persistent Volume when opting to deploy
+ClickHouse with Persistent Volumes. This choice is particularly crucial because
+if you deploy ClickHouse with Persistent Volumes (PV) and use the default
+ZooKeeper, there's a risk that if the ZooKeeper pod crashes before the ClickHouse
+pod, ClickHouse will automatically switch the table to a READONLY state. This
+sequential occurrence can lead to repeated crashes of the ClickHouse pod because
+all data stored in ZooKeeper is lost in the event of a ZooKeeper crash.
```
yanjunz97, line 306: I'm thinking about maybe we do not need to include that many technical details?
Like just saying 

```
It's important to provision ZooKeeper with Persistent Volume when opting to deploy
ClickHouse with Persistent Volumes, as the crashing of in-memory ZooKeeper will lead
to data lost, with which the ClickHouse will keep crashing.
```

Please feel free to rephrasing it.

build/charts/theia/values.yaml
```diff
@@ -150,7 +146,40 @@ clickhouse:
...
       securityContext:
         runAsUser: 1000
-        fsGroup: 1000 
+        fsGroup: 1000
+      storage:
+        # -- Zookeeper storage size. Can be a plain integer or as a fixed-point
+        # number using one of these quantity suffixes: E, P, T, G, M, K. Or the
+        # power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki.
+        size: "5Gi"
+        createPersistentVolume:
```
yanjunz97, line 155: Do you think we need to also enable ZooKeeper PV in the e2e when we enable PV for ClickHouse? (I think only the migrate_clickhouse_test is using ClickHouse with PV for now. Please tell me if I miss any other test.)

yuntanghsu: We currently use ClickHouse with PV for `migrate_clickhouse_test` and `upgrade_test`.
I think it's safer to enable ZooKeeper PV when enabling PV for ClickHouse.
Do you think we need to use a separate version? Like only enable ZooKeeper for version higher than v0.8.0? Or I can just enable it whenever --local flag is used?

docs/network-flow-visibility.md
yuntanghsu: Fixed. Thanks

docs/network-flow-visibility.md
yuntanghsu: Fixed. Thanks.

docs/network-flow-visibility.md
yuntanghsu: Updated.

build/charts/theia/values.yaml
yuntanghsu: I also found an issue that when upgrading the ZooKeeper from default one to the one with PV, the data will be completely lost, leading ClickHouse to keep crashing again.

docs/network-flow-visibility.md
```diff
@@ -286,10 +286,22 @@ will lead to a complete data loss. In that case, please refer to the
 to recover the data.
 
 ClickHouse cluster can be deployed with default Local PV or NFS PV by setting
-`clickhouse.storage.createPersistentVolume`. To have more flexibility in the
-PV creation, you can configure a customized `StorageClass` in
+`clickhouse.storage.createPersistentVolume`. To have more flexibility in the PV
+creation, you can configure a customized `StorageClass` in
 `clickhouse.storage.persistentVolumeClaimSpec`.
 
+From Theia 0.8, We also offer the option of deploying Zookeeper with Persistent
```
yanjunz97, line 293:
```suggestion
From Theia 0.8, we also offer the option of deploying Zookeeper with Persistent
```

build/charts/theia/values.yaml
yanjunz97: We discussed offline and agreed that we could enable ZooKeeper PV in e2e tests in next release to avoid the upgrading issue between in-memory and PV

yanjunz97 (approved):
Overall LGTM, only a nit

tushartathgur (approved):
lgtm

PR #489, opened 2023-09-27
Remove uninitialized taint to recover CAPV testbeds

XinShuYang:
The updated CAPV cluster will generate uninitialized taint on worker nodes. To avoid blocking the tests, we need to update related script.


yuntanghsu (approved):
LGTM

PR #493, opened 2023-09-28
Add PodWaitForReady for ClickHouse Pod in waitForClickHousePod

yuntanghsu:
Previously, we only check if CurrentRevision is changed to UpdateRevision in waitForClickHousePod when applying a new Flow Visibility manifest. It is possible that the StatefulSet updates the CurrentRevision before the new pod is ready. To make the e2e test robust, we add PodWaitForReady for ClickHouse Pod in waitForClickHousePod.

### DIFF
--- a/test/e2e/framework.go
+++ b/test/e2e/framework.go
@@ -168,8 +168,9 @@ const (
 )
 
 var (
-	errConnectionLost = fmt.Errorf("http2: client connection lost")
-	clickHousePodName = fmt.Sprintf("%s-0-0-0", clickHousePodNamePrefix)
+	errConnectionLost         = fmt.Errorf("http2: client connection lost")
+	clickHousePodName         = fmt.Sprintf("%s-0-0-0", clickHousePodNamePrefix)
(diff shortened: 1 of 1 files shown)


test/e2e/framework.go
```diff
@@ -1410,13 +1410,17 @@ func (data *TestData) waitForClickHousePod() error {
 		// The updating is completed when CurrentRevision is updated to UpdateRevision
 		return ss.Status.CurrentRevision == ss.Status.UpdateRevision, nil
 	})
+	clickHousePodName := fmt.Sprintf("%s-0-0-0", clickHousePodNamePrefix)
```
tushartathgur, line 1413: Please use the const mentioned as L172. Could you change at L1323 also?

yuntanghsu: Thanks for reviewing. I've fixed it and also moved clickHouseStatefulSetName to the same place.

dreamtalen (approved):
LGTM

tushartathgur (approved):
LGTM

PR #536, opened 2023-10-10
Migrate to create-plugin and Update package version for grafana-sankey-plugin and grafana-chord-plugin

yuntanghsu:
In this PR we do:
1. Migration from toolkit to create-plugin for grafana-sankey-plugin and grafana-chord-plugin.
2. Update node version as create-plugin doesn't support node-version < 18.
3. Update @grafana/ui and @grafana/data for grafana-sankey-plugin and grafana-chord-plugin.
4. Update the unit test for grafana-sankey-plugin and grafana-chord-plugin.

Reviewer can focus on files [ package.json/jest.config.js/ChordPanel.test.tsx/ChordPanel.tsx/SankeyPanel.tsx]

Note:
1. Couple of dependencies are updated by create-plugin automatically.
2. .config file is generated by create-plugin automatically.
3. Do not update the version of "react" and "react-dom" to the version higher than 16, as we are using enzyme-adapter-react-16. Using version higher than 16 will break the unit test and show error `Cannot read property 'child' of undefined`
4. The reason we add transformIgnorePatterns in jest.config.js is because we are not using grafana/toolkit to run the test and Jest failed to parse a file. This happens e.g. when our code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

Granafa pictures:
Screenshot 2023-10-13 at 1 24 30 PM
Screenshot 2023-10-13 at 1 24 43 PM
Screenshot 2023-10-13 at 1 25 57 PM
Screenshot 2023-10-13 at 1 26 16 PM
Screenshot 2023-10-13 at 1 26 27 PM
Screenshot 2023-10-13 at 1 26 37 PM
Screenshot 2023-10-13 at 1 33 10 PM
Screenshot 2023-10-13 at 1 33 26 PM


plugins/grafana-custom-plugins/grafana-chord-plugin/src/ChordPanel.test.tsx
```diff
@@ -41,14 +42,26 @@ describe('Chord Panel Test', () => {
...
-        const arcSpy = jest.spyOn(d3, 'arc');
-        const ribbonSpy = jest.spyOn(d3, 'ribbonArrow');
-        const scaleOrdinalSpy = jest.spyOn(d3, 'scaleOrdinal');
-        const wrapper = mount(<UnthemedChordPanel {...props} />);
-
-        expect(wrapper.find(UnthemedChordPanel).exists()).toEqual(true);
-        expect(arcSpy).toHaveBeenCalled();
-        expect(ribbonSpy).toHaveBeenCalled();
-        expect(scaleOrdinalSpy).toHaveBeenCalled();
+        const { unmount } = render(<UnthemedChordPanel {...props} />);
```
heanlan, line 45: The naming of `unmount` is bit confusing to me. Could you help explain why we call `unmount()` in L64 will do the clean up?

plugins/grafana-custom-plugins/grafana-chord-plugin/src/ChordPanel.tsx
```diff
@@ -8,14 +8,17 @@ interface Props extends Themeable2, PanelProps<ChordOptions> {}
...
   const d3Container = useRef(null);
-
   // GetFieldVal reads field's value from data series
   function GetFieldVal(fieldName: string) {
     return data.series
       .map((series: DataFrame) => series.fields.find((field: any) => field.name.toLowerCase() === fieldName.toLowerCase()))
       .map((field: any) => {
         let record = field?.values as any;
-        return record?.buffer;
+        if(Array.isArray(record)) {
```
heanlan, line 17: Out of curiosity, can we just do `let record = field.display!(field.values)`? I'm not sure whether it will run into issues, just referencing https://github.com/grafana/grafana/blob/16d0aff26707b80fda408708d1c67bb2fe0586e9/packages/grafana-ui/src/components/Table/utils.ts#L521
The `else` block is a bit confusing for me. Does it mean if the record is not an array, then we should just return null?

yuntanghsu: The reason I check if record is array is that I found that record is different type when I use jest and webpack.
For webpack, it is a vectorArray and record?.buffer is an array. But for jest, record is array already and record?.buffer will be "undefined", which will break the test.

plugins/grafana-custom-plugins/grafana-chord-plugin/src/ChordPanel.test.tsx
yuntanghsu: That is a feature when we use @testing-library/react.
https://testing-library.com/docs/react-testing-library/api/#render-result

The unmount function returned by the render function in the @testing-library/react library is a function that you can call to unmount (remove) the React component from the DOM and perform necessary cleanup.

In React components that use the useEffect hook, we can specify a cleanup function as the return value of the useEffect callback. This cleanup function is executed when the component is unmounted.

```
useEffect(() => {
  // Effect code

  return () => {
    // Cleanup code (executed when the component is unmounted)
  };
}, [dependencies]);
```

plugins/grafana-custom-plugins/grafana-chord-plugin/.config/Dockerfile
```diff
@@ -0,0 +1,16 @@
+ARG grafana_version=latest
```
Dhruv-J, line 1: should we use latest version or verified version? (I don't think this file matters that much to be honest, but maybe we could match it with the grafana version we use in our deployment as well?)

Dhruv-J (commented):
overall LGTM, just some nits

build/charts/theia/README.md
```diff
@@ -66,7 +66,7 @@ Kubernetes: `>= 1.16.0-0`
...
 | grafana.image | object | `{"pullPolicy":"IfNotPresent","repository":"projects.registry.vmware.com/antrea/theia-grafana","tag":"9.1.6"}` | Container image used by Grafana. |
-| grafana.installPlugins | list | `["https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-sankey-plugin-1.0.2.zip;theia-grafana-sankey-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-chord-plugin-1.0.1.zip;theia-grafana-chord-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-dependency-plugin-1.0.2.zip;theia-grafana-dependency-plugin","grafana-clickhouse-datasource 2.2.0"]` | Grafana plugins to install. |
+| grafana.installPlugins | list | `["https://github.com/yuntanghsu/grafana_sankey_plugin/archive/refs/tags/v1.0.7.zip;theia-grafana-sankey-plugin","https://github.com/yuntanghsu/grafana_chord_plugin/archive/refs/tags/v1.0.8.zip;theia-grafana-chord-plugin","https://downloads.antrea.io/artifacts/grafana-custom-plugins/theia-grafana-dependency-plugin-1.0.2.zip;theia-grafana-dependency-plugin","grafana-clickhouse-datasource 2.2.0"]` | Grafana plugins to install. |
```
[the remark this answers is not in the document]
yuntanghsu: For test purpose. Will change it to final version after this PR is apporved.

plugins/grafana-custom-plugins/grafana-chord-plugin/.config/Dockerfile
yuntanghsu: I've removed this file as we don't need it. Thanks!

plugins/grafana-custom-plugins/grafana-chord-plugin/src/ChordPanel.tsx
heanlan: could you help try if `let record = field.display!(field.values)` will work for both jest and webpack?

If not, could you add a comment in the code for the above explanation?

plugins/grafana-custom-plugins/grafana-chord-plugin/.config/Dockerfile
heanlan: I thought the files under `./config` are all automatically generated. How do we distinguish which files are needed and which are not?

plugins/grafana-custom-plugins/grafana-chord-plugin/src/ChordPanel.tsx
yuntanghsu: That doesn't work for jest.
```
field.display is not a function
TypeError: field.display is not a function
```
I've added comments to explain the reason. Please take a look to see if it is clear enough?

plugins/grafana-custom-plugins/grafana-chord-plugin/.config/Dockerfile
yuntanghsu: Do we need Dockerfile to build the image for grafana? I think this file is redundant even it is auto-generated?

heanlan (approved):
LGTM

PR #540, opened 2023-10-12
Fix vulneralibilities found in snowflake golangci

dreamtalen:


snowflake/pkg/utils/file/file.go
```diff
@@ -78,7 +80,7 @@ func DownloadAndUntar(ctx context.Context, logger logr.Logger, url string, dir s
 		if err != nil {
 			return err
 		}
-		dest := filepath.Join(dir, hdr.Name)
+		dest := filepath.Join(dir, filepath.Clean(hdr.Name))
```
tushartathgur, line 83: I don't think clean will help as per the example here https://www.geeksforgeeks.org/filepath-clean-function-in-golang-with-examples/,
we should eliminate any ".." with current path. Please correct me if I am wrong

snowflake/pkg/utils/file/file.go
```diff
@@ -29,6 +29,8 @@ import (
 	"github.com/go-logr/logr"
 )
 
+const maxCopySize = 100 * 1024 * 1024 // Limit to 100 MB
```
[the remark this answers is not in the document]
dreamtalen: It's not a standard but rather a quick fix for the `G110: Potential DoS vulnerability via decompression bomb (gosec)` in snowflake golangci. 
We can expand it when we have the opportunity to do e2e testing.

snowflake/pkg/utils/file/file.go
dreamtalen: Thanks for pointing out. Yes Clean() can't eliminate all ".." in the path, but I believe we are not expected to change the path in this case.
Therefore, I added an error return when ".." is still detected after Clean().

tushartathgur: I guess the new check will be sufficient.

tushartathgur (approved):
These measures should be enough to not have a vulnerability.
LGTM

PR #557, opened 2023-10-16
Change ClickHouse volume ownership

yanjunz97:


elton-furtado (approved):

yanjunz97:
> Thanks Yanjun for takeing care this. Do you think we will have similar issue for ZooKeeper or Grafana? I couldn't reproduce the issue neither. Not sure what happened.

ZooKeeper and Grafana should be fine.

I remember we have similar work on grafana side for this at https://github.com/antrea-io/theia/blob/63156479ed45c86e600552d0d9f053efd39ec54a/build/charts/theia/templates/grafana/deployment.yaml#L24-L30
As for zookeeper, I also find some code here which seems to take care of the folder ownership issue
https://github.com/antrea-io/theia/blob/63156479ed45c86e600552d0d9f053efd39ec54a/build/charts/theia/templates/clickhouse/zookeeper/statefulset.yaml#L135-L136

yuntanghsu (approved):
LGTM

PR #592, opened 2023-10-27
Revert "Bump aquasecurity/trivy-action from 0.12.0 to 0.13.0"

tushartathgur:
Reverts antrea-io/theia#578


yuntanghsu:
Why are we reverting this?

